PR #6, opened 2019-02-12
Remove pods when workflow is finished

schrodit:
Adds the feature to configure if all pods of a workflow should be deleted when the workflow itself has finished.

This feature can be toggled by setting the environment variable "CLEAN_WORKFLOW_PODS"

**Release note**:

```improvement operator
Adds configurable cleanup of pods when workflow has finished
```

### DIFF
--- a/docs/Testmachinery.md
+++ b/docs/Testmachinery.md
@@ -1,0 +1,16 @@
+# Testmachinery
+
+
+## Controller
+
+| Name                | default                                                                 | Description                                                                                               |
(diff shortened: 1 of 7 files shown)


dguendisch:
@schrodit Does that preserve the logs or are they gone after pod deletion as well?

dguendisch:
Just see that you commented in the documentation that logs are preserved even on pod deletion.
Guess that answers my question :)

PR #15, opened 2019-02-25
Add support for automatic determination of k8s version

schrodit:
- Add dynamic determination of k8s as default if no version provided in the testrunner
- refactor testrunner cli to use cobra lib with automatic required checks

Fixes #

**Release note**:

```improvement user
Refactor Testrunner cli to use cobra
```
```improvement user
Add automatic k8s version to Testrunner if no value is defined
```


pkg/testrunner/helper.go
```diff
@@ -0,0 +1,68 @@
...
+			versions[i] = v
+		}
+	}
+	sort.Sort(semver.Collection(versions))
+
+	return versions[len(versions)-1].String(), nil
+}
+
+func getCloudproviderVersions(profile *gardenv1beta1.CloudProfile, cloudprovider string) ([]string, error) {
+	switch gardenv1beta1.CloudProvider(cloudprovider) {
```
dguendisch: Is it possible to fetch the version information in a "generic" way? One where we don't have to adapt the testrunner once a new provider is added?

schrodit: unfortunately not because we need a mapping from a gostruct field to a string.

I think it could be generic by using reflection, but this has a bad smell and I think we should be consistent with the gardener style (https://github.com/gardener/gardener/blob/master/pkg/operation/cloudbotanist/cloudbotanist.go#L46)

docs/GetStarted.md
```diff
@@ -21,11 +21,12 @@ A TestDefinition consists of a name and a description of how the test should be
 Therefore, a basic test consists of a command which will be executed inside the specified container image.
 
 ```yaml
+kind: TestDefinition
 metadata:
   name: TestDefName
 spec:
-  owner: gardener@team.de # test owner and contact person in case of a test failure 
-  recipientsOnFailure: developer1@team.de, developer2@team.de # optional, list of emails to be notified if a step fails 
+  owner: gardener@team.de # test owner and contact person in case of a test failure
```
dguendisch:
```suggestion
  owner: gardener@example.com # test owner and contact person in case of a test failure
```

docs/GetStarted.md
```diff
@@ -21,11 +21,12 @@ A TestDefinition consists of a name and a description of how the test should be
...
+  recipientsOnFailure: developer1@team.de, developer2@team.de # optional, list of emails to be notified if a step fails
```
dguendisch:
```suggestion
  recipientsOnFailure: developer1@example.com, developer2@example.com # optional, list of emails to be notified if a step fails
```

PR #22, opened 2019-03-01
Refactor testrunner to make it more modular

schrodit:
This PR refactors the testrunner and transforms it into a more usable cli.
- testrunner stages are now modularized to use them more or less independent
  - Stages: Generate Helm Template, run testruns, collect and store results, notify
  - Better testing as modules are independent
- a new command is added to run a specific testrun so no more need for a helm template 

**Release note**:

```improvement user
Add testrunner command to run specific testrun without a helm template and output, notification
```
```improvement user
Make poll interval configurable.
```
```action user
Default poll interval changed to 20 seconds. Previous default was 60 seconds.
```


schrodit:
The PR is not in WIP anymore. It can be reviewed

schrodit:
reviewed offline

PR #23, opened 2019-03-04
Split conformance into serial and parallel tests

schrodit:
Split conformance tests into serial and parallel tests.

**Release note**:

```improvement user
Split conformance tests into serial and parallel
```

### DIFF
--- a/.test-defs/ConformanceSerial.yaml
+++ b/.test-defs/ConformanceSerial.yaml
@@ -14,7 +14,7 @@
 
 kind: TestDefinition
 metadata:
-  name: conformance
+  name: conformance-serial
 spec:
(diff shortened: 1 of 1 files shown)


.test-defs/ConformanceSerial.yaml
```diff
@@ -0,0 +1,36 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+kind: TestDefinition
+metadata:
+  name: conformance-serial
+spec:
+  owner: oleg.loewen@sap.com
+  recipientsOnFailure:
+  - oleg.loewen@sap.com
```
dguendisch: How about using DL Gardener Output Qualification here?

schrodit: Done

PR #29, opened 2019-03-08
Add "ingested" status to a testruns status

schrodit:
This PR adds an attribute to a testruns status which indicates if the testrun is already processed and the  results are ingested.
This status will be also updated by testrunner when it's finished ingesting the results to elasticsearch.

**Release note**:

```improvement operator
NONE
```

### DIFF
--- a/cmd/testrunner/cmd/runtestrun/run_testrun.go
+++ b/cmd/testrunner/cmd/runtestrun/run_testrun.go
@@ -96,7 +96,7 @@ func init() {
 	runTestrunCmd.Flags().StringVar(&tmKubeconfigPath, "tm-kubeconfig-path", "", "Path to the testmachinery cluster kubeconfig")
 	runTestrunCmd.MarkFlagRequired("tm-kubeconfig-path")
 	runTestrunCmd.MarkFlagFilename("tm-kubeconfig-path")
-	runTestrunCmd.Flags().StringVarP(&namespace, "namespace", "n", "default", "Namesapce where the testrun should be deployed.")
+	runTestrunCmd.Flags().StringVarP(&namespace, "namespace", "n", "default", "Namespace where the testrun should be deployed.")
 
 	runTestrunCmd.Flags().Int64Var(&timeout, "timeout", 3600, "Timout in seconds of the testrunner to wait for the complete testrun to finish.")
 	runTestrunCmd.Flags().Int64Var(&interval, "interval", 20, "Poll interval in seconds of the testrunner to poll for the testrun status.")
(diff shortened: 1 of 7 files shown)


cmd/testrunner/cmd/runtemplate/run_template.go
```diff
@@ -162,6 +164,7 @@ func init() {
 	runCmd.Flags().StringVar(&elasticSearchConfigName, "es-config-name", "sap_internal", "The elasticsearch secret-server config name.")
 	runCmd.Flags().StringVar(&s3Endpoint, "s3-endpoint", os.Getenv("S3_ENDPOINT"), "S3 endpoint of the testmachinery cluster.")
 	runCmd.Flags().StringVar(&concourseOnErrorDir, "concourse-onError-dir", os.Getenv("ON_ERROR_DIR"), "On error dir which is used by Concourse.")
+	runCmd.Flags().BoolVar(&updateIngest, "update-ingest", true, "Update the status of a ingested testrun.")
```
dguendisch: Maybe that's more clear?
```suggestion
	runCmd.Flags().BoolVar(&updateIngest, "update-ingest", true, "Update the ingestion result (success/failure) of a testrun.")
```

docs/testrunner/testrunner_run-template.md
```diff
@@ -38,6 +38,7 @@ testrunner run-template [flags]
       --testruns-chart-path string         Path to the testruns chart.
       --timeout int                        Timout in seconds of the testrunner to wait for the complete testrun to finish. (default 3600)
       --tm-kubeconfig-path string          Path to the testmachinery cluster kubeconfig
+      --update-ingest                      Update the status of a ingested testrun. (default true)
```
dguendisch: Maybe this is more clear?

```suggestion
      --update-ingest                      Update the ingestion result (success/failure) of a testrun. (default true)
```

cmd/testrunner/cmd/runtemplate/run_template.go
schrodit: Maybe we should rename it completely to a more general term as maybe there are more status updates in the future?
How about `update-status`?

dguendisch: Good idea. On a second thought: is there a reason why to not update the status? In which cases would one want to have the status not updated? So maybe we should always update and remove that option completely?

schrodit: yes true I could remove this flag and apply it as soon as the ingest function successfully finishes.
My thought was to add this possibility for later commands (get results of an already run testrun) . But yes it is really not needed

schrodit: Removed this flag

docs/testrunner/testrunner_run-template.md
schrodit: see above

PR #31, opened 2019-03-14
Add ssl support for s3 instances

schrodit:
Adds support for ssl secured s3 instances.
Also a new flag is introduced `s3-ssl` to specify ssl usage of the corresponding endpoint.

**Release note**:

```improvement operator
Add ssl support for s3 instances
```

### DIFF
--- a/pkg/testrunner/result/output.go
+++ b/pkg/testrunner/result/output.go
@@ -40,7 +40,6 @@ func Output(config *Config, tmKubeconfigPath, namespace string, tr *tmv1beta1.Te
 	if config.OutputFile == "" {
 		return nil
 	}
-	var s3Endpoint = config.S3Endpoint
 
 	metadata.TestrunID = tr.Name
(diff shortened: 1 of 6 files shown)


pkg/testrunner/result/output.go
```diff
@@ -59,7 +59,7 @@ func Output(config *Config, tmKubeconfigPath, namespace string, tr *tmv1beta1.Te
 		return err
 	}
 
-	osConfig, err := getOSConfig(tmKubeconfigPath, s3Endpoint, namespace)
+	osConfig, err := getOSConfig(tmKubeconfigPath, s3Endpoint, namespace, config.S3SSL)
```
dguendisch: you could inline `s3Endpoint` to `config.S3Endpoint` and remove the var on line 43

schrodit: done

PR #33, opened 2019-03-15
Add additional phase for timeouts

schrodit:
This PR adds an additional phase called "Timeout" which applied if a argo node exceeds their `activeDeadlineSeconds`.

Should we add the `finishedAt` and `duration` according to the activeDeadlineSeconds so that finishedAt = startTime + duration and duration = activeDeadlineSeconds.

This would mean that we have to add additional information to the status (activeDeadlineSeconds of the TestDefinitiobn) as not all information of the TestDefinition is available in the status.
**Release note**:

```improvement user
Add additional phase "Timeout" for test that exceeded the `activeDeadlineSeconds`
```


schrodit:
Added the duration to the status of the testrun so that now the "real" duration is shown in the status and in elasticsearch

pkg/controller/testrun_control_update.go
```diff
@@ -104,6 +108,17 @@ func updateStepsStatus(tr *tmv1beta1.Testrun, wf *argov1.Workflow) {
 				continue
 			}
 
+			if strings.Contains(argoNodeStatus.Message, argoctrl.ErrDeadlineExceeded.Error()) {
+				completionTime := metav1.NewTime(stepStatus.StartTime.Add(time.Duration(*stepStatus.TestDefinition.ActiveDeadlineSeconds) * time.Second))
```
OlegLoewen: Maybe decouple a little bit for better reading?
```suggestion
				testDurationMs := time.Duration(*stepStatus.TestDefinition.ActiveDeadlineSeconds) * time.Second
				completionTime := metav1.NewTime(stepStatus.StartTime.Add(testDurationMs))
```

pkg/testmachinery/testdefinition/testdefinition.go
```diff
@@ -52,13 +52,16 @@ func New(def *tmv1beta1.TestDefinition, loc Location, fileName string) *TestDefi
 	if def.Spec.Image == "" {
 		def.Spec.Image = testmachinery.BASE_IMAGE
 	}
+	if def.Spec.ActiveDeadlineSeconds == nil {
+		def.Spec.ActiveDeadlineSeconds = &activeDeadlineSeconds
```
OlegLoewen:
```suggestion
		def.Spec.ActiveDeadlineSeconds = &defaultActiveDeadlineSeconds
```

pkg/util/util.go
```diff
@@ -45,7 +45,7 @@ func MaxTimeExceeded(startTime time.Time, maxWaitTime int64) bool {
 
 // Completed checks if the testrun is in a completed phase
 func Completed(phase argov1.NodePhase) bool {
-	if phase == argov1.NodeSucceeded || phase == argov1.NodeFailed || phase == argov1.NodeError || phase == argov1.NodeSkipped {
+	if phase == tmv1beta1.PhaseStatusSuccess || phase == tmv1beta1.PhaseStatusFailed || phase == tmv1beta1.PhaseStatusError || phase == tmv1beta1.PhaseStatusSkipped || phase == tmv1beta1.PhaseStatusTimeout {
```
OlegLoewen: Optional/Idea: as discussed maybe it's good to introduce a list.contains utility function to make this line more compact and readable? (In combination with a dot import of tmv1beta1)
```suggestion
    if stringInSlice(phase, []string{PhaseStatusSuccess, PhaseStatusFailed, PhaseStatusError, PhaseStatusSkipped, PhaseStatusTimeout}) {
func stringInSlice(a string, list []string) bool {
    for _, b := range list {
        if b == a {
            return true
        }
    }
    return false
}
```

pkg/testmachinery/testdefinition/testdefinition.go
schrodit: done

pkg/controller/testrun_control_update.go
schrodit: done

pkg/util/util.go
schrodit: Discussed offline

schrodit:
@OlegLoewen @dguendisch added an integration for the new phase and fixed dependency errors.
can you please rereview?

schrodit:
Problem was I needed another package for the pod message which resulted in a need to update some additional packages.
But I try to separate them next time:)

PR #35, opened 2019-03-18
Rename result output attribute names and update Docs

schrodit:
Rename multiple metadata fields:
-  `tm_meta` to `tm`
- `kubernetes_version` to `k8s_version`
- `bom` to `component_descriptor` (only updated in internal structure, json output stays `bom`)

**Release note**:

```action user
Renamed metadata fields: `tm_meta` to `tm` and `kubernetes_version` to `k8s_version`.
```


schrodit:
> Not sure about the `bom` to `component_descriptor` renaming, recall that we can't rename fields in kibana, so `component_descriptor` looks really long. I think `bom` is fine enough and well known enough.
> The other renamings are fine.

Ok, just thought that we should be consistent as this is the only place were we call the component descriptor bom.

schrodit:
@dguendisch updated bom name

PR #36, opened 2019-03-19
Add argoui link to metadata output

schrodit:
This PR adds another optional metadata field which is a link that points to the wf in an argo ui.
Therefore another flag `argoui-endpoint` is introduced where the argoui endpoint can be specified which is then used to generate the URL for the specific workflow.

**Release note**:

```improvement user
Add ArgoUI link to metadata.
```

### DIFF
--- a/pkg/testrunner/result/output.go
+++ b/pkg/testrunner/result/output.go
@@ -23,6 +23,8 @@ import (
 	"fmt"
 	"io"
 	"io/ioutil"
+	"net/url"
+	"path"
 	"strings"
 
 	"github.com/gardener/test-infra/pkg/testmachinery"
@@ -44,6 +46,10 @@ func Output(config *Config, tmClient kubernetes.Interface, namespace string, tr 
 	}
 
 	metadata.TestrunID = tr.Name
+
+	if _, err := url.ParseRequestURI(config.ArgoUIEndpoint); err != nil {
+		metadata.ArgoUIExternalURL = path.Join(config.ArgoUIEndpoint, "workflows", namespace, tr.Name)
+	}
 
 	trSummary, err := getTestrunSummary(tr, metadata)
 	if err != nil {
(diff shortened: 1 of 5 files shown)


pkg/testrunner/result/types.go
```diff
@@ -57,7 +60,14 @@ type Metadata struct {
...
 	ComponentDescriptor []*componentdescriptor.Component `json:"bom"`
-	TestrunID           string                           `json:"testrun_id"`
+
+	// Name of the testrun crd object.
+	TestrunID string `json:"testrun_id"`
+
+	// Link to the workflow in the argo ui
+	// Only set if an ingress with the name "argo-ui" for the argoui is set.
+	// Argo URL is in format: https://argo-endpoint.com/workflows/namespace/wf-name
+	ArgoUIExternalURL string `json:"argo_external_url"`
```
dguendisch: Again for the kibana reason: can we shorten the field name? (would then also match `cc_meta:concourse_url`)
```suggestion
	ArgoUIExternalURL string `json:"argo_url"`
```

schrodit: sure :+1:
I wasn't which url field to use as "inspiration" cause there is also this field: `cc_meta.atc-external-url`

dguendisch: You are right and the justification by pointing to another existing field is actually pointless, it's all about having short but still meaningful field names :)

pkg/testrunner/result/output.go
```diff
@@ -45,6 +47,10 @@ func Output(config *Config, tmClient kubernetes.Interface, namespace string, tr
 
 	metadata.TestrunID = tr.Name
 
+	if _, err := url.ParseRequestURI(config.ArgoUIEndpoint); err != nil {
```
OlegLoewen: Shouldn't this be...? Not sure. Also the error message isn't handled. I'm missing something like: if err != nil { logger.log("Can't parse ArgoUIEndpoint")...
```suggestion
	if _, err := url.ParseRequestURI(config.ArgoUIEndpoint); err == nil {
```

schrodit: I wanted to use this check just to verify if the url is set and if it is a valid url.
If not then the field should just be ignored and only added in case of no error

schrodit: done

PR #41, opened 2019-03-25
Add shared folder to artifacts

OlegLoewen:
Adds shared folder which can be used across the flow steps.

**Release note**:

```improvement operator
Add shared folder for testflow steps.
```

### DIFF
--- a/docs/GetStarted.md
+++ b/docs/GetStarted.md
@@ -4,6 +4,7 @@
   - [Create a Test](#create-a-test)
     - [Input Contract](#input-contract)
     - [Export Contract](#export-contract)
+    - [Shared Folder](#shared-folder)
     - [Images](#images)
   - [Create a Testrun](#create-a-testrun)
   - [Configuration](#configuration)
@@ -116,6 +117,10 @@ It then automatically uploads these documents to an index named like the TestDef
       { "index": { "_index": "mySecondSpecificIndex", "_type": "_doc" } }
       { "key3": 5 }
     ```
+
+### Shared Folder
+
+Data that is stored in `TM_SHARED_PATH` location, can be accessed from within any testflow step of a the workflow. This is essential if e.g. a test flow step needs to evaluate the output of the previously finished test flow step. This folder is also available as an artifact in the Argo UI.
 
 ### Images
 
--- a/pkg/testmachinery/types.go
+++ b/pkg/testmachinery/types.go
@@ -24,7 +24,10 @@ import (
 
 const (
 	// TM_KUBECONFIG_PATH is the path where kubeconfigs are mounted to tests.
-	TM_KUBECONFIG_PATH = "/tmp/env/kubeconfig"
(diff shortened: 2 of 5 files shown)


pkg/testmachinery/types.go
```diff
@@ -26,6 +26,9 @@ const (
 	// TM_KUBECONFIG_PATH is the path where kubeconfigs are mounted to tests.
 	TM_KUBECONFIG_PATH = "/tmp/env/kubeconfig"
 
+	// TM_KUBECONFIG_PATH is the path where kubeconfigs are mounted to tests.
+	TM_SHARED_PATH = "/tmp/shared"
```
schrodit: I would change this to a testmachinery namespace --> `tmp/tm/shared`

schrodit: don't know why we had TM_KUBECONFIG_PATH at `tmp/env/kubeconfig`, maybe you could change this path as well to `tmp/tm/kubeconfig`

OlegLoewen: Done

docs/GetStarted.md
```diff
@@ -117,6 +118,10 @@ It then automatically uploads these documents to an index named like the TestDef
       { "key3": 5 }
     ```
 
+### Shared Folder
+
+Data that is stored in `TM_SHARED_PATH` location, can be accessed from within any testflow step of a the workflow. This is essential if e.g. a test flow step needs to evaluate the output of the previously finished test flow step. This folder also available as an artifact in the Argo UI.
```
dguendisch:
```suggestion
Data that is stored in `TM_SHARED_PATH` location, can be accessed from within any testflow step of a the workflow. This is essential if e.g. a test flow step needs to evaluate the output of the previously finished test flow step. This folder is also available as an artifact in the Argo UI.
```

pkg/testmachinery/types.go
```diff
@@ -24,7 +24,10 @@ import (
 
 const (
 	// TM_KUBECONFIG_PATH is the path where kubeconfigs are mounted to tests.
-	TM_KUBECONFIG_PATH = "/tmp/env/kubeconfig"
+	TM_KUBECONFIG_PATH = "/tmp/tm/kubeconfig"
+
+	// TM_KUBECONFIG_PATH is the path where kubeconfigs are mounted to tests.
```
dguendisch: should be an own comment (copy/paste) :)

OlegLoewen: Done

test/controller/testflow/testflow_suite_test.go
```diff
@@ -177,6 +178,54 @@ var _ = Describe("Testflow execution tests", func() {
 		})
 	})
 
+	Context("sharedFolder", func() {
+		sharedFilePath := fmt.Sprintf("%s/%s", testmachinery.TM_SHARED_PATH, "test")
```
schrodit: move to test if not needed in other tests

PR #48, opened 2019-04-03
added kibana log url

dguendisch:
Adds a direct link to the metadata of each testrun and teststep pointing to the respective kibana logging installation with a an active filter on the specific testrun.

### DIFF
--- a/cmd/testrunner/cmd/runtestrun/run_testrun.go
+++ b/cmd/testrunner/cmd/runtestrun/run_testrun.go
@@ -90,6 +90,7 @@ var runTestrunCmd = &cobra.Command{
 		run := []*testrunner.Run{
 			{
 				Testrun: &tr,
+				Metadata: &testrunner.Metadata{},
 			},
 		}
 
--- a/pkg/testrunner/result/output.go
+++ b/pkg/testrunner/result/output.go
@@ -19,24 +19,24 @@ import (
 	"bytes"
 	"compress/gzip"
 	"context"
-	"encoding/json"
 	"fmt"
 	"io"
 	"io/ioutil"
 	"net/url"
 	"path"
 	"strings"
-
-	"github.com/gardener/test-infra/pkg/testrunner"
+	"text/template"
 
 	"github.com/gardener/test-infra/pkg/testmachinery"
+	"github.com/gardener/test-infra/pkg/testrunner"
 	"github.com/gardener/test-infra/pkg/testrunner/elasticsearch"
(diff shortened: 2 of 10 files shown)


pkg/testrunner/result/output.go
```diff
@@ -268,3 +276,23 @@ func getOSConfig(tmClient kubernetes.Interface, namespace, minioEndpoint string,
 		BucketName: minioConfig.Data["objectstore.bucketName"],
 	}, nil
 }
+
+// buildKibanaWorkflowURL construct a valid kibana url from the given endpoint and workflow
+// typical endpoint: https://kibana.ingress.<my-cluster-domain>
+// typical kibana url: https://kibana.ingress.<my-cluster-domain>/app/kibana#/...
+func buildKibanaWorkflowURL(endpointPath string, namespace string, workflow string) string {
+	const IndexPatternID = "d1134060-5189-11e9-b7dc-3fcc67c9fd25"
```
schrodit: nit: lowercase variable name `indexPatternID`

pkg/util/util.go
```diff
@@ -169,3 +170,16 @@ func PrettyPrintStruct(obj interface{}) string {
 	}
 	return string(str)
 }
+
+// MarshalNoHTMLEscape is nearly same as json.Marshal but does NOT HTLM-escape <, > or &
+// However it does add a newline char at the end (as done by json.Encoder.Encode)
+func MarshalNoHTMLEscape(v interface{}) ([]byte, error) {
```
schrodit: Maybe include JSON in the function name to specify that this method Mashals json data.

cmd/testrunner/cmd/runtestrun/run_testrun.go
```diff
@@ -90,6 +90,7 @@ var runTestrunCmd = &cobra.Command{
 		run := []*testrunner.Run{
 			{
 				Testrun: &tr,
+				Metadata: &testrunner.Metadata{},
```
schrodit: Thanks for the fix;)

pkg/testrunner/result/output.go
dguendisch: you're right, changed it (confused it with the uppercase requirement for the respective struct for json marshalling)

pkg/util/util.go
dguendisch: as discussed per slack, leaving it like that for now

dguendisch:
applied changes, ptal

PR #50, opened 2019-04-04
Add testrunner runtime metadata to testrun annotations

schrodit:
Add metadata information to the testrun as annotations.

We need check how the bom can be persisted (maybe use a configmap)
**Release note**:

```improvement user
Add runtime metadata into annotations of a testrun.
```


pkg/testrunner/template/template.go
```diff
@@ -27,6 +27,7 @@ import (
 )
 
 // Render renders a helm chart with containing testruns, adds the provided parameters and values, and returns the parsed and modified testruns.
+// Adds the compont descriptor to metadata.
```
OlegLoewen:
```suggestion
// Adds the component descriptor to metadata.
```

PR #51, opened 2019-04-08
moved kibana url types to exclude from custom exported metadata

dguendisch:
Moves kibana url type definitions from broad Metadata to specific TestrunSummary and StepSummary to avoid "polluting" all custom exported metadata with this "huge" URL.
Adds specific kibana urls to StepSummary so that we can present a direct link to the logs of a specific TestStep.

**Release note**:

```improvement user
Adds kibana URLs to elasticsearch metadata to directly link to the logs of a respective TestRun or TestStep.
```

### DIFF
--- a/pkg/testrunner/result/ingest.go
+++ b/pkg/testrunner/result/ingest.go
@@ -35,7 +35,7 @@ const (
 func IngestDir(path string, esCfgName string) error {
 	files, err := ioutil.ReadDir(path)
 	if err != nil {
-		return fmt.Errorf("Cannot read directory %s: %s", path, err.Error())
+		return fmt.Errorf("cannot read directory '%s'd: %s", path, err.Error())
 	}
 	for _, file := range files {
 		if !file.IsDir() {
--- a/pkg/testrunner/result/output.go
+++ b/pkg/testrunner/result/output.go
@@ -51,15 +51,7 @@ func Output(config *Config, tmClient kubernetes.Interface, namespace string, tr 
 		}
 	}
 
-	if config.KibanaEndpoint != "" && tr.Status.Workflow != "" {
-		if u, err := url.ParseRequestURI(config.KibanaEndpoint); err == nil {
-			metadata.KibanaExternalURL = buildKibanaWorkflowURL(u.String(), namespace, tr.Status.Workflow)
-		} else {
-			log.Debugf("Cannot parse kibana Url %s: %s", config.KibanaEndpoint, err.Error())
-		}
-	}
-
-	trSummary, err := getTestrunSummary(tr, metadata)
+	trSummary, err := getTestrunSummary(tr, metadata, config)
 	if err != nil {
 		return err
 	}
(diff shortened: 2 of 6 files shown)


schrodit:
Please rebase to not include output-dir-path changes

dguendisch:
This PR is currently ahead of master by one commit, nothing to rebase.

pkg/testrunner/result/ingest.go
```diff
@@ -35,7 +35,7 @@ const (
 func IngestDir(path string, esCfgName string) error {
 	files, err := ioutil.ReadDir(path)
 	if err != nil {
-		return fmt.Errorf("Cannot read directory %s: %s", path, err.Error())
+		return fmt.Errorf("cannot read directory '%s'd: %s", path, err.Error())
```
schrodit: nit: Cannot/cannot (gland displays and error but i think we should be consistent in our error messages)

pkg/testrunner/result/output.go
```diff
@@ -240,22 +245,31 @@ func getFilesFromTar(r io.Reader) ([][]byte, error) {
...
+		log.Errorf("cannot create kibana template url: %s", err.Error())
 	}
+
 	builder := strings.Builder{}
 	builder.WriteString(endpointPath)
+	filters := kibanaFilter{indexPatternID, workflow, testStepName}
 	err = tmpl.Execute(&builder, filters)
 	if err != nil {
-		log.Errorf("cannot template kibana url with ns '%s' and workflow '%s': %s", namespace, workflow, err.Error())
+		log.Errorf("cannot template kibana url for workflow '%s' and testStepName '%s': %s", workflow, testStepName, err.Error())
```
schrodit: nit Cannot

pkg/testrunner/result/output.go
```diff
@@ -118,6 +108,13 @@ func getTestrunSummary(tr *tmv1beta1.Testrun, metadata *testrunner.Metadata) ([]
 				StartTime: step.StartTime,
 				Duration:  step.Duration,
 			}
+			if config.KibanaEndpoint != "" && tr.Status.Workflow != "" && summary.Name != "" {
```
schrodit: Maybe include all/some checks in the buildKibanaLogURL function to keep the getTestrunSummary lean?

pkg/testrunner/result/ingest.go
dguendisch: Yes, I know that we use upper-case so far, but I'd follow https://github.com/golang/go/wiki/CodeReviewComments#error-strings
Either we can change them over time as we change nearby code (this is what I did), or change them in one shot everywhere, or leave them upper-case (which I think would be the least good option).
WDYT?

pkg/testrunner/result/output.go
dguendisch: Makes absolutely sense, looks much better now, thanks for the hint!

dguendisch:
Changed and commented, PTAL

pkg/testrunner/result/ingest.go
schrodit: your're right. We should change them over time.
Maybe we can include this when doing the error refactoring

pkg/testrunner/result/output.go
schrodit: looks good now:)

schrodit:
> This PR is currently ahead of master by one commit, nothing to rebase.
> You probably refer to an output-dir-path change I added to the collector command (it was missing there)

Thanks for the fix;)

PR #52, opened 2019-04-08
Split bulk into files of a specific size

schrodit:
Split bulk into files of 50 MB.

**Release note**:

```improvement user
Split bulk files into files of 50MB
```
```action user
`output-file-path` parameter has changed to `output-dir-path`
```


pkg/testrunner/elasticsearch/elasticsearchbulk.go
```diff
@@ -24,27 +24,61 @@ import (
...
 
-var newLine = []byte("\n")
+//var newLine = []byte("\n")
 
-// Marshal creates a elastic search bulk json of its metadata and sources.
-func (b *Bulk) Marshal() (*bytes.Buffer, error) {
+// 50 mb
+const maxBufferSize = 50 * 1024 * 1024
+
+// Marshal creates a elastic search bulk json of its metadata and sources; and returns a list of bulk files with a max size of 4mb
```
dguendisch:
```suggestion
// Marshal creates an elastic search bulk json of its metadata and sources and returns a list of bulk files with a max size of 50mb
```

pkg/testrunner/elasticsearch/elasticsearchbulk.go
```diff
@@ -24,27 +24,61 @@ import (
...
+	buf.Write(meta)
+	buf.Write(b.Source)
+
+	return buf.Bytes(), nil
+}
 
-	for _, source := range b.Sources {
-		buf.Write(meta)
-		buf.Write(source)
+// NewList creates an list of Bulks with the same metadata
```
dguendisch:
```suggestion
// NewList creates a list of Bulks with the same metadata
```

pkg/testrunner/result/ingest.go
```diff
@@ -30,6 +32,22 @@ const (
 	cliPath = "/cc/utils/cli.py"
 )
 
+func IngestDir(path string, esCfgName string) error {
+	files, err := ioutil.ReadDir(path)
+	if err != nil {
+		return fmt.Errorf("Cannot read direactory %s: %s", path, err.Error())
```
dguendisch:
```suggestion
		return fmt.Errorf("Cannot read directory %s: %s", path, err.Error())
```

PR #61, opened 2019-04-18
Add disk attach performance test

schrodit:
Adds a integration performance test for volumes.
The test follows this workflow:
- create 10 statefulsets with a nginx and vm's of 1Gi
- delete the machine in the seed so that mcm drains the node and creates a new one
- for every single sts:
  - wait until the pods support unreadiness to have the real downtime of such a service
  - start time measurement
  - wait for pods to become healthy again
  - Stop measurement
- report all results back into a separate es index

**Release note**:

```improvement operator
Add a performance test for disk attach and detach operations
```


test/k8s-tests/perf-disk-attach/disk_attach_test.go
```diff
@@ -0,0 +1,256 @@
...
+	. "github.com/gardener/gardener/test/integration/framework"
+	. "github.com/gardener/gardener/test/integration/shoots"
+
+	appsv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+var (
+	InitializationTimeout = 1800 * time.Second
```
dguendisch: Maybe `30 * time.Minute` is easier to read :)

OlegLoewen:
Does it make sense, that you explain this changes in person?

PR #70, opened 2019-04-26
Fix result collection

schrodit:
Fix marking of persited tr.

**Release note**:

```improvement operator
NONE
```


pkg/testrunner/result/helper.go
```diff
@@ -31,11 +32,11 @@ func writeBulks(path string, bufs [][]byte) error {
...
 
-func writeToFile(fielPath string, data []byte) error {
-
-	err := ioutil.WriteFile(fielPath, data, 0644)
+func writeToFile(filePath string, data []byte) error {
+	log.Debugf("write to file %s", filePath)
+	err := ioutil.WriteFile(filePath, data, 0644)
 	if err != nil {
-		return fmt.Errorf("Cannot write to %s", err.Error())
+		return fmt.Errorf("cannot write to %s", err.Error())
```
dguendisch:
```suggestion
		return fmt.Errorf("cannot write to '%s': %s", filePath, err.Error())
```

PR #81, opened 2019-06-04
expose gardener version

OlegLoewen:
Exposes Gardener Version

**Release note**:

```improvement operator
expose gardener version
```

### DIFF
--- a/pkg/testrunner/template/render.go
+++ b/pkg/testrunner/template/render.go
@@ -51,7 +51,9 @@ func renderSingleChart(renderer chartrenderer.ChartRenderer, parameters *Testrun
 			"autoscalerMax":        parameters.AutoscalerMax,
 			"floatingPoolName":     parameters.FloatingPoolName,
 			"loadbalancerProvider": parameters.LoadBalancerProvider,
-			"gardenerVersion":      parameters.GardenerVersion,
+			"gardener": map[string]interface{}{
+				"version": parameters.GardenerVersion,
+			},
 		},
 		"kubeconfigs": map[string]interface{}{
 			"gardener": string(gardenKubeconfig),
--- a/pkg/testrunner/template/render.go
+++ b/pkg/testrunner/template/render.go
@@ -51,9 +51,9 @@ func renderSingleChart(renderer chartrenderer.ChartRenderer, parameters *Testrun
 			"autoscalerMax":        parameters.AutoscalerMax,
 			"floatingPoolName":     parameters.FloatingPoolName,
 			"loadbalancerProvider": parameters.LoadBalancerProvider,
-			"gardener": map[string]interface{}{
-				"version": parameters.GardenerVersion,
-			},
+		},
+		"gardener": map[string]interface{}{
+			"version": parameters.GardenerVersion,
 		},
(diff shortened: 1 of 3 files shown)


pkg/testrunner/template/template.go
```diff
@@ -83,3 +84,12 @@ func Render(tmClient kubernetes.Interface, parameters *TestrunParameters, metada
 
 	return testruns, nil
 }
+
+func exposeGardenerVersionToParameters(componentDescriptor componentdescriptor.ComponentList, parameters *TestrunParameters) {
+	for _, component := range componentDescriptor {
+		if component.Name == "github.com/gardener/gardener.version" {
+			parameters.GardenerVersion = component.Version
+			break
```
schrodit: I would rather return instead of breaking out of the loop.
```suggestion
			return
```

pkg/testrunner/template/render.go
```diff
@@ -52,6 +51,7 @@ func renderSingleChart(renderer chartrenderer.ChartRenderer, parameters *Testrun
 			"autoscalerMax":        parameters.AutoscalerMax,
 			"floatingPoolName":     parameters.FloatingPoolName,
 			"loadbalancerProvider": parameters.LoadBalancerProvider,
+			"gardenerVersion":      parameters.GardenerVersion,
```
schrodit: I would create a new object in the helm vallues file for the gardener version.
So that it will be at `.Values.gardener.version` instead of being in the shoot values `.Values.shoot.gardenerVersion`

pkg/testrunner/template/types.go
```diff
@@ -38,6 +38,7 @@ type TestrunParameters struct {
 	FloatingPoolName        string
 	LoadBalancerProvider    string
 	ComponentDescriptorPath string
+	GardenerVersion         string
```
schrodit: nit: add a new line to separate this value from the other normal parameters

pkg/testrunner/template/render.go
OlegLoewen: done

pkg/testrunner/template/types.go
OlegLoewen: done

pkg/testrunner/template/render.go
```diff
@@ -52,6 +51,9 @@ func renderSingleChart(renderer chartrenderer.ChartRenderer, parameters *Testrun
 			"autoscalerMax":        parameters.AutoscalerMax,
 			"floatingPoolName":     parameters.FloatingPoolName,
 			"loadbalancerProvider": parameters.LoadBalancerProvider,
+			"gardener": map[string]interface{}{
```
schrodit: just move this one level up:
```go
"shoot": map[string]interface{}{
		"name":                 fmt.Sprintf("%s-%s", parameters.ShootName, util.RandomString(5)),
		"projectNamespace":     fmt.Sprintf("garden-%s", parameters.ProjectName),
                ....
},
"gardener": map[string]interface{}{
		"version": parameters.GardenerVersion,
},
```

OlegLoewen: done

schrodit: thanks

PR #91, opened 2019-06-11
introduce node.UseGlobalArtifacts

OlegLoewen:
introduce node.UseGlobalArtifacts

**Release note**:

```
Add support for the usage of GlobalArtifacts in steps
```

### DIFF
--- a/pkg/apis/testmachinery/v1beta1/types.go
+++ b/pkg/apis/testmachinery/v1beta1/types.go
@@ -234,7 +234,7 @@ type TestFlow []*DAGStep
 type DAGStep struct {
 	Name               string         `json:"name,omitempty"`
 	Definition         StepDefinition `json:"definition,omitempty"`
-	UseGlobalArtifacts bool           `json:"useGlobalInput,omitempty"`
+	UseGlobalArtifacts bool           `json:"useGlobalArtifacts,omitempty"`
 	DependsOn          []string       `json:"dependsOn,omitempty"`
 	ArtifactsFrom      string         `json:"artifactsFrom,omitempty"`
 }
--- a/pkg/apis/testmachinery/v1beta1/types.go
+++ b/pkg/apis/testmachinery/v1beta1/types.go
@@ -232,10 +232,11 @@ type ConfigElement struct {
 type TestFlow []*DAGStep
 
 type DAGStep struct {
-	Name          string         `json:"name,omitempty"`
-	Definition    StepDefinition `json:"definition,omitempty"`
-	DependsOn     []string       `json:"dependsOn,omitempty"`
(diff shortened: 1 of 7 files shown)


pkg/apis/testmachinery/v1beta1/types.go
```diff
@@ -232,10 +232,11 @@ type ConfigElement struct {
 type TestFlow []*DAGStep
 
 type DAGStep struct {
-	Name          string         `json:"name,omitempty"`
-	Definition    StepDefinition `json:"definition,omitempty"`
-	DependsOn     []string       `json:"dependsOn,omitempty"`
-	ArtifactsFrom string         `json:"artifactsFrom,omitempty"`
+	Name               string         `json:"name,omitempty"`
+	Definition         StepDefinition `json:"definition,omitempty"`
+	UseGlobalArtifacts bool           `json:"useGlobalInput,omitempty"`
```
schrodit: Adjust json name

pkg/testmachinery/testflow/node/node.go
```diff
@@ -169,6 +147,54 @@ func (n *Node) Status() *tmv1beta1.StepStatus {
 	return status
 }
 
+func (n *Node) isRootNode() bool {
+	return n.inputSource != nil
+}
+
+func (n *Node) GlobalArtifacts() []argov1.Artifact {
```
schrodit: `GlobalArtifacts` is misleading.
We should name something like `generateArtifacts` or `getTaksArtifacts`

pkg/testmachinery/testflow/node/node.go
```diff
@@ -169,6 +147,54 @@ func (n *Node) Status() *tmv1beta1.StepStatus {
...
 }
 
+func (n *Node) isRootNode() bool {
+	return n.inputSource != nil
+}
+
+func (n *Node) DetermineArtifacts() []argov1.Artifact {
+	artifactsMap := make(map[string]argov1.Artifact)
+
+	if n.step.UseGlobalArtifacts {
```
schrodit: if we use maps then we could move the global artifact down and remove the if in line 168.
Then we can get rid of one if condition and have a more semantic code that global artifacts are not the default.
```go
if n.isRootNode() {
  artifactMap["kubeconfig"]...
  ...
}
if n.step.UseGlobalArtifacts {
  ...
}
```

pkg/testmachinery/testflow/node/node.go
OlegLoewen: done

pkg/testmachinery/testflow/node/node.go
OlegLoewen: done

PR #92, opened 2019-06-12
ArtifactsFrom

OlegLoewen:
Add artifactsFrom feature

Fixes #92 
****Release note**:

```
Add artifactsFrom feature

```

### DIFF
--- a/pkg/testmachinery/testflow/node/node.go
+++ b/pkg/testmachinery/testflow/node/node.go
@@ -25,8 +25,6 @@ import (
 	"github.com/gardener/test-infra/pkg/testmachinery/testdefinition"
 )
 
-var artifacts []argov1.Artifact
-
 // CreateNodesFromStep creates new nodes from a step and adds default configuration
 func CreateNodesFromStep(step *tmv1beta1.DAGStep, loc locations.Locations, globalConfig []*config.Element, flowID string) (Set, error) {
 	testdefinitions, err := loc.GetTestDefinitions(step.Definition)
@@ -154,9 +152,7 @@ func (n *Node) isRootNode() bool {
 }
 
 func (n *Node) GetOrDetermineArtifacts() []argov1.Artifact {
-	if artifacts != nil {
-		return artifacts
-	}
+	var artifacts []argov1.Artifact
 	artifactsMap := make(map[string]argov1.Artifact)
 	if n.isRootNode() {
 		artifactsMap["kubeconfigs"] = argov1.Artifact{
--- a/pkg/testmachinery/testflow/node/node.go
+++ b/pkg/testmachinery/testflow/node/node.go
@@ -158,10 +158,6 @@ func (n *Node) GetOrDetermineArtifacts() []argov1.Artifact {
 		return artifacts
 	}
 	artifactsMap := make(map[string]argov1.Artifact)
-	if n.step.ArtifactsFrom != "" {
-		artifacts = n.getPreviousNode(n.step.ArtifactsFrom).GetOrDetermineArtifacts()
-		return artifacts
-	}
 	if n.isRootNode() {
 		artifactsMap["kubeconfigs"] = argov1.Artifact{
 			Name: "kubeconfigs",
--- a/pkg/testmachinery/testflow/node/node.go
+++ b/pkg/testmachinery/testflow/node/node.go
@@ -25,6 +25,8 @@ import (
 	"github.com/gardener/test-infra/pkg/testmachinery/testdefinition"
 )
 
+var artifacts []argov1.Artifact
+
 // CreateNodesFromStep creates new nodes from a step and adds default configuration
 func CreateNodesFromStep(step *tmv1beta1.DAGStep, loc locations.Locations, globalConfig []*config.Element, flowID string) (Set, error) {
 	testdefinitions, err := loc.GetTestDefinitions(step.Definition)
@@ -108,7 +110,7 @@ func (n *Node) Name() string {
 
 // Task returns the argo task definition for the node.
 func (n *Node) Task() argov1.DAGTask {
-	task := argo.CreateTask(n.TestDefinition.GetName(), n.TestDefinition.GetName(), testmachinery.PHASE_RUNNING, n.step.Definition.ContinueOnError, n.ParentNames(), n.DetermineArtifacts())
+	task := argo.CreateTask(n.TestDefinition.GetName(), n.TestDefinition.GetName(), testmachinery.PHASE_RUNNING, n.step.Definition.ContinueOnError, n.ParentNames(), n.GetOrDetermineArtifacts())
 
 	switch n.step.Definition.Condition {
 	case tmv1beta1.ConditionTypeSuccess:
@@ -151,8 +153,15 @@ func (n *Node) isRootNode() bool {
 	return n.inputSource != nil
 }
 
-func (n *Node) DetermineArtifacts() []argov1.Artifact {
+func (n *Node) GetOrDetermineArtifacts() []argov1.Artifact {
+	if artifacts != nil {
+		return artifacts
+	}
 	artifactsMap := make(map[string]argov1.Artifact)
+	if n.step.ArtifactsFrom != "" {
+		artifacts = n.getPreviousNode(n.step.ArtifactsFrom).GetOrDetermineArtifacts()
+		return artifacts
+	}
 	if n.isRootNode() {
 		artifactsMap["kubeconfigs"] = argov1.Artifact{
 			Name: "kubeconfigs",
@@ -181,7 +190,25 @@ func (n *Node) DetermineArtifacts() []argov1.Artifact {
 			From: "{{workflow.outputs.artifacts.sharedFolder}}",
 		}
 	}
-	return artifactsMapToList(artifactsMap)
+	artifacts = artifactsMapToList(artifactsMap)
+	return artifacts
+}
+
+func (n *Node) getPreviousNode(previousNodeStepName string) *Node {
+	if n.Parents == nil || len(n.Parents) == 0 {
+		return nil
+	}
+	for parent := range n.Parents {
+		if parent.step.Name == previousNodeStepName {
+			return parent
+		}
+	}
+	for parent := range n.Parents {
+		if previousNode := parent.getPreviousNode(previousNodeStepName); previousNode != nil {
+			return previousNode
+		}
+	}
+	return nil
 }
 
 func artifactsMapToList(m map[string]argov1.Artifact) []argov1.Artifact {
--- a/pkg/testmachinery/testflow/validation.go
+++ b/pkg/testmachinery/testflow/validation.go
@@ -54,7 +54,7 @@ func Validate(identifier string, tf tmv1beta1.TestFlow, locs locations.Locations
 
 	usedTestdefinitions := 0
 
-	usedStepNames := make(map[string]bool, 0)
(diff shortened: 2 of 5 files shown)


schrodit:
Please rebase

pkg/testmachinery/testflow/validation.go
```diff
@@ -99,7 +107,45 @@ func Validate(identifier string, tf tmv1beta1.TestFlow, locs locations.Locations
 	return nil
 }
 
-func checkIfDependentStepsExist(identifier string, tf tmv1beta1.TestFlow, usedStepNames map[string]bool) error {
+func ensureArtifactsFromStepsExist(identifier string, tf tmv1beta1.TestFlow, usedStepNames map[string]*tmv1beta1.DAGStep) error {
+	for stepName, step := range usedStepNames {
```
schrodit: The identifier in line 112 is only used for the user to print out errors and show exactly where the error occurs.
But here the user doenst know anything about our internal list.

Therefor I would loop through the the testflow list so we can point to the exact erroneous step.
```suggestion
	for i, step := range tf {
```

pkg/testmachinery/testflow/validation.go
```diff
@@ -99,7 +107,45 @@ func Validate(identifier string, tf tmv1beta1.TestFlow, locs locations.Locations
...
+		identifier := fmt.Sprintf("%s.[%s]", identifier, stepName)
```
schrodit:
```suggestion
		identifier := fmt.Sprintf("%s.%d", identifier, i)
```

pkg/testmachinery/testflow/validation.go
```diff
@@ -99,7 +107,45 @@ func Validate(identifier string, tf tmv1beta1.TestFlow, locs locations.Locations
...
+		if step.ArtifactsFrom != "" {
+			if !previousStepExists(step, step.ArtifactsFrom, usedStepNames) {
+				return fmt.Errorf("%s.artifactsFrom: Invalid value: Step %s is unknown", identifier, step.ArtifactsFrom)
```
schrodit:
```suggestion
				return fmt.Errorf("%s: Invalid value: Step %s is unknown", identifier, step.ArtifactsFrom)
```

pkg/testmachinery/testflow/validation.go
```diff
@@ -99,7 +107,45 @@ func Validate(identifier string, tf tmv1beta1.TestFlow, locs locations.Locations
...
+			}
+		}
+	}
+	return nil
+}
+
+func previousStepExists(step *tmv1beta1.DAGStep, previousStepName string, usedStepNames map[string]*tmv1beta1.DAGStep) bool {
+	// no need to check if ArtifactsFrom is empty
+	if step.ArtifactsFrom == "" {
```
schrodit: remove, as this is an recursive function it would return false for a parent with artifactsfrom.
But a parent without artifacts can still be a valid parent

pkg/testmachinery/testflow/validation.go
```diff
@@ -99,7 +107,45 @@ func Validate(identifier string, tf tmv1beta1.TestFlow, locs locations.Locations
...
+		return false
+	}
+
+	if step.DependsOn == nil || len(step.DependsOn) == 0 {
+		// if there are no dependents, then there can be no matching ArtifactsFrom step
+		return false
+	} else {
```
schrodit: no need for the else branch as the condition before returns if true.
So just remove it

pkg/testmachinery/testflow/validation.go
```diff
@@ -99,7 +107,45 @@ func Validate(identifier string, tf tmv1beta1.TestFlow, locs locations.Locations
...
+		// first check if one of direct dependents is the ArtifactsFrom step
+		for _, dependentStep := range step.DependsOn {
```
schrodit: nit: `dependentStep` / `dependentStepName`

pkg/testmachinery/testflow/node/node.go
```diff
@@ -151,8 +153,15 @@ func (n *Node) isRootNode() bool {
...
 }
 
-func (n *Node) DetermineArtifacts() []argov1.Artifact {
+func (n *Node) GetOrDetermineArtifacts() []argov1.Artifact {
+	if artifacts != nil {
+		return artifacts
+	}
 	artifactsMap := make(map[string]argov1.Artifact)
+	if n.step.ArtifactsFrom != "" {
+		artifacts = n.getPreviousNode(n.step.ArtifactsFrom).GetOrDetermineArtifacts()
```
schrodit: this is not correct, there you would get the input artifacts of the àrtifactsFrom`step but you want the output artifacts of the previous step.

Therefore I would suggest to move this functionality to this method: https://github.com/gardener/test-infra/blob/2b67714452faaa9de0a03e12192c13f3dd4e0ef4/pkg/testmachinery/testflow/flow_operations.go#L135.
Then you can just set the `artifactsFrom` artifact as we already do it with the default input artifacts.
```go
func ApplyOutputScope(steps map[string]*Step) error {
	for _, step := range steps {
		for n := range step.Nodes {
			var outputSourceNode *node.Node
			if step.ArtifactsFrom != "" {
				outputSourceNode = steps[step.ArtifactsFrom]
			} else {
				parents := n.Parents
				for len(parents) != 1 {
					parents = parents.GetParents()
					if len(parents) == 0 {
						return fmt.Errorf("no serial parent node can be found for step %s in node %s", n.Name(), step.Info.Name)
					}
				}
				outputSourceNode = parents.List()[0]
			}
			outputSourceNode.EnableOutput()
			n.SetInputSource(outputSourceNode)
		}
	}

	return nil
}
```

OlegLoewen: okay, by chance, do you mean: `outputSourceNode = steps[step.Info.ArtifactsFrom].Nodes.List()[0]` instead of `outputSourceNode = steps[step.ArtifactsFrom]`

schrodit: ah yes your're right

OlegLoewen: done

pkg/testmachinery/testflow/validation.go
OlegLoewen: done

pkg/testmachinery/testflow/validation.go
OlegLoewen: done

pkg/testmachinery/testflow/validation.go
OlegLoewen: done

OlegLoewen:
Please recheck

pkg/testmachinery/testflow/validation.go
```diff
@@ -99,6 +107,55 @@ func Validate(identifier string, tf tmv1beta1.TestFlow, locs locations.Locations
...
+		if step.ArtifactsFrom != "" {
+			if !previousStepExists(step, step.ArtifactsFrom, tf) {
+				return fmt.Errorf("%s; Invalid value: Couldn't find artifactsFrom step %s", identifier, step.ArtifactsFrom)
+			}
+		}
+	}
+	return nil
+}
+
+func getStep(tf tmv1beta1.TestFlow, needle string) *tmv1beta1.DAGStep {
```
schrodit: this can be removed if you add `usedStepNames` to the parameters of `ensureArtifactsFromStepsExist` and `previousStepExists`.

Then you simply pass a pointer around and can get the step as before with `usedStepNames[dependentStepName] `

pkg/testmachinery/testflow/validation.go
```diff
@@ -99,6 +107,55 @@ func Validate(identifier string, tf tmv1beta1.TestFlow, locs locations.Locations
 	return nil
 }
 
+func ensureArtifactsFromStepsExist(identifier string, tf tmv1beta1.TestFlow) error {
```
schrodit:
```suggestion
func ensureArtifactsFromStepsExist(identifier string, tf tmv1beta1.TestFlow, steps map[string]*node.Node) error {
```

pkg/testmachinery/testflow/validation.go
```diff
@@ -99,6 +107,55 @@ func Validate(identifier string, tf tmv1beta1.TestFlow, locs locations.Locations
...
+	}
+	for _, step := range tf {
+		if step.Name == needle {
+			return step
+		}
+	}
+	return nil
+}
+
+func previousStepExists(step *tmv1beta1.DAGStep, previousStepName string, tf tmv1beta1.TestFlow) bool {
```
schrodit:
```suggestion
func previousStepExists(step *tmv1beta1.DAGStep, previousStepName string, tf tmv1beta1.TestFlow, steps map[string]*node.Node) bool {
```

pkg/testmachinery/testflow/validation.go
```diff
@@ -99,6 +107,55 @@ func Validate(identifier string, tf tmv1beta1.TestFlow, locs locations.Locations
...
+
+	// first check if one of direct dependents is the ArtifactsFrom step
+	for _, dependentStepName := range step.DependsOn {
+		if dependentStepName == previousStepName {
+			return true
+		}
+	}
+	// second check if dependents of dependents has the ArtifactsFrom step
+	for _, dependentStepName := range step.DependsOn {
+		if previousStepExists(getStep(tf, dependentStepName), previousStepName, tf) {
```
schrodit:
```suggestion
		if previousStepExists(steps[dependentStepName]), previousStepName, tf) {
```

pkg/testmachinery/testflow/validation.go
OlegLoewen: done

pkg/testmachinery/testflow/validation.go
OlegLoewen: done

pkg/testmachinery/testflow/validation.go
OlegLoewen: done

pkg/testmachinery/testflow/validation.go
OlegLoewen: done

pkg/testmachinery/testflow/validation.go
OlegLoewen: done

OlegLoewen:
Please recheck

pkg/testmachinery/testflow/node/node.go
```diff
@@ -151,7 +153,10 @@ func (n *Node) isRootNode() bool {
 	return n.inputSource != nil
 }
 
-func (n *Node) DetermineArtifacts() []argov1.Artifact {
+func (n *Node) GetOrDetermineArtifacts() []argov1.Artifact {
+	if artifacts != nil {
```
schrodit: where does artifacts come from?

please have a look at the test results

pkg/testmachinery/testflow/node/node.go
```diff
@@ -25,6 +25,8 @@ import (
 	"github.com/gardener/test-infra/pkg/testmachinery/testdefinition"
 )
 
+var artifacts []argov1.Artifact
```
schrodit: global var will lead to unknown behavior as the controller is executed concurrently

OlegLoewen: done

pkg/testmachinery/testflow/node/node.go
OlegLoewen: done

PR #120, opened 2019-07-30
Add node affinity & tolerations to workflow pods

dguendisch:
Reverses the recently added affinity/tolerations feature, so that workflow pods specify a node preference. This results in having only workflow pods (except daemonsets of course) running on dedicated nodes.
Workflow pods now prefer nodes labeled and tainted with `purpose: testload`.

Fixes #

**Release note**:
```improvement operator
Workflow pods (aka testload) prefer now nodes that are tainted/labeled with `purpose: testload`.
This reverses and thus removes the previous approach of prefering specific nodes only for testmachinery components (controllers, etc).
```

### DIFF
--- a/pkg/testmachinery/argo/argo.go
+++ b/pkg/testmachinery/argo/argo.go
@@ -17,12 +17,10 @@ package argo
 import (
 	argov1 "github.com/argoproj/argo/pkg/apis/workflow/v1alpha1"
 	argoclientset "github.com/argoproj/argo/pkg/client/clientset/versioned"
+	"github.com/gardener/test-infra/pkg/testmachinery"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/client-go/tools/clientcmd"
 )
-
-// the label and taint of the nodes in the worker pool which are reserved for workflow pods
-const workerPoolTaintLabelName = "testload"
 
 // CreateWorkflow takes a name, templates and volumes to generate an argo workflow object.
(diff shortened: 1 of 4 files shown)


pkg/testmachinery/argo/argo.go
```diff
@@ -21,11 +21,16 @@ import (
 	"k8s.io/client-go/tools/clientcmd"
 )
 
+// the label and taint of the nodes in the worker pool which are reserved for workflow pods
+const workerPoolTaintLabelName = "testload"
```
schrodit: I would move this constant to https://github.com/gardener/test-infra/blob/a939f23ebc0c12cb3018788bea385f860cd91505/pkg/testmachinery/config.go
then we have all configuration in one place and do not need to search for configuration in different places.

schrodit: Maybe also make it configurable from the outside via parameter or env var like [here](https://github.com/gardener/test-infra/blob/7b2ed5f6a3b78dd7de3b1e84bf74e6557eac62f9/pkg/testmachinery/testmachinery.go#L32)

dguendisch: I moved the config.
I guess there's no need for making this feature configurable as workflow pods only _prefer_ nodes labeled with `purpose: testload`. If no such nodes exist, the workflow pods will happily schedule to whatever nodes are available.

schrodit: ok then this is ok

PR #121, opened 2019-07-30
Move hostscheduler code to pkg

schrodit:

- moves the hostscheduler test code to pkg to keep the cmd directories clean and have all business logic in `pkg``
- Only cleanup the cluster if the cluster is not yet freed

**Release note**:

```improvement operator
NONE
```


pkg/testmachinery/argo/argo.go
```diff
@@ -102,7 +102,7 @@ func getWorkflowAffinity() *corev1.Affinity {
 		NodeAffinity: &corev1.NodeAffinity{
 			PreferredDuringSchedulingIgnoredDuringExecution: []corev1.PreferredSchedulingTerm{
 				{
-					Weight:     100,
+					Weight: 100,
```
dguendisch: Hmm, do we have different formatters?

schrodit: ah yeah could be. I use `go fmt`

dguendisch: I think goland is correct in this case. Could it be that you run an old version of gofmt? See e.g. https://github.com/golang/go/issues/26497

dguendisch: Well, just checked the latest gofmt (1.12.7) and it also wrongly formats the above...

dguendisch: Nevermind, then it is like it is and we stick to go fmt :)

PR #122, opened 2019-08-02
Use go-modules

schrodit:
Use go modules

**Release note**:

```improvement operator
TestMachinery now use go modules
```


schrodit:
> Even better and with go 1.13 we can get of `GO111MODULE...` as well.

and with 1.14 we can also get rid of the `-mod=vendor`..

PR #132, opened 2019-08-23
For all-k8s-versions run only for newest patch version

OlegLoewen:
If all-k8s-versions is used, run only tests for newest patch version, instead of considering all versions that are listed in cloud profiles.

Too many TM tests are triggered in testgrid testruns

**Release note**:

```
If all-k8s-versions used, run only testruns for newest patch version
```


schrodit:
closed in favour of https://github.com/gardener/test-infra/pull/133

PR #143, opened 2019-09-10
Pin kubetest to specific version

schrodit:

Fixes #

**Release note**:

```improvement operator
Kubetest is now pinned to a specific commit 
```

### DIFF
--- a/integration-tests/e2e/config/config.go
+++ b/integration-tests/e2e/config/config.go
@@ -59,6 +59,10 @@ var (
 	RunCleanUpAfterTest      bool
 )
 
+const (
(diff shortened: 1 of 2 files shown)


integration-tests/e2e/config/config.go
```diff
@@ -59,6 +59,10 @@ var (
 	RunCleanUpAfterTest      bool
 )
 
+const (
+	TestInfraVersion = "466f8ffd615df63e539e96c4c713959958f620e5"
```
dguendisch:
```suggestion
	// points to a commit in https://github.com/kubernetes/test-infra/commits/master
	TestInfraVersion = "466f8ffd615df63e539e96c4c713959958f620e5"
```

PR #146, opened 2019-09-13
Fix insecure flag in controller deployment.

jkbschmid:
Fix missing '-' for local controller deployment.

**Release note**:

```NONE

```


dguendisch:
Thanks @jkbschmid. Can be merged after you signed the CLA.

PR #156, opened 2019-09-30
Attach testrun status as github release asset

OlegLoewen:
Attach testrun status as github release asset, for easier reporting.

**Release note**:

```
Attach testrun status as github release asset, for easier reporting.
```


pkg/testrunner/result/collect.go
```diff
@@ -63,6 +65,27 @@ func (c *Collector) Collect(log logr.Logger, tmClient kubernetes.Interface, name
 			}
 		}
 
+		// upload testrun status to github component
+		if cfg.GithubComponentForStatus != "" {
```
schrodit: maybe we should be more explicit and add an `UploadStatusAsset bool` parameter.
WDYT?

pkg/testrunner/result/collect.go
```diff
@@ -63,6 +65,27 @@ func (c *Collector) Collect(log logr.Logger, tmClient kubernetes.Interface, name
...
+			if cfg.GithubPassword == "" || cfg.GithubUser == "" || cfg.ComponentDescriptorPath == "" {
+				runLogger.Error(err, "missing github password / github user / component descriptor path argument")
+			}
+			components, err := componentdescriptor.GetComponentsFromFile(cfg.ComponentDescriptorPath)
```
schrodit: handle error

pkg/testrunner/result/collect.go
schrodit: maybe we can move this check into a separate function as proposed by you?
now I see that it is more code than I thought so it would make sense to factor it out.
Then we can also get rid of these else nesting

pkg/testrunner/result/status-uploader.go
```diff
@@ -0,0 +1,113 @@
...
+	tmv1beta1 "github.com/gardener/test-infra/pkg/apis/testmachinery/v1beta1"
+)
+
+// uploads status results as asset to the component
+func UploadStatusToGithub(run *testrunner.Run, component *componentdescriptor.Component, githubUser, githubPassword string) error {
+	tr := run.Testrun
+	md := run.Metadata
+	tableString := strings.Builder{}
+	util.RenderStatusTable(&tableString, tr.Status.Steps)
+	statusOutput := fmt.Sprintf("Testrun: %s\n%s\n%s", tr.GenerateName, tableString.String(), util.PrettyPrintStruct(tr.Status))
```
schrodit: use `tr.Name` as this is the real unique name to identify the testrun

pkg/testrunner/result/status-uploader.go
```diff
@@ -0,0 +1,113 @@
...
+	filename := fmt.Sprintf("%s-%s-%s.txt", md.Landscape, md.CloudProvider, md.KubernetesVersion)
+	if err := ioutil.WriteFile(filename, []byte(statusOutput), 0644); err != nil {
```
schrodit: it seems weird that we need to write the file first then reopen to upload it.
I think we should use `NewUploadRequest` so we can directly pass the `io.Reader`

pkg/testrunner/result/status-uploader.go
```diff
@@ -0,0 +1,113 @@
...
+	// check if asset exists and delete if so
+	releaseAssets, response, err := githubClient.Repositories.ListReleaseAssets(context.Background(), repoOwner, repoName, *release.ID, &github.ListOptions{Page: 1, PerPage: 200})
+	if err != nil {
+		return err
+	} else if response.StatusCode != 200 {
+		return errors.New(fmt.Sprintf("Get github release assets failed with status code %d", response.StatusCode))
+	}
+	for _, releaseAsset := range releaseAssets {
+		if *releaseAsset.Name == filename {
+			response, err = githubClient.Repositories.DeleteReleaseAsset(context.Background(), repoOwner, repoName, *releaseAsset.ID)
```
schrodit: Shouldn't we jsut edit the asset instead of delete and reupload?

pkg/apis/testmachinery/v1beta1/types.go
```diff
@@ -143,7 +143,10 @@ type TestrunStatus struct {
 	Steps []*StepStatus `json:"steps,omitempty"`
 
 	// Ingested states whether the result of a testrun is already ingested into a persistant storage (db).
-	Ingested bool `json:"ingested"`
+	Ingested         bool `json:"ingested"`
+
+	// UploadedToGithub states whether the status of a testrun is already uploaded to github component
+	UploadedToGithub bool `json:"uploadedToGithub"`
```
schrodit: make it `*bool` as it is optional and omit if empty

pkg/testrunner/result/status-uploader.go
OlegLoewen: Editing seems to much more complex. I'm even not sure if editing the content is possible, but only the metadata.

pkg/testrunner/result/status-uploader.go
OlegLoewen: Can you elaborate?

pkg/testrunner/result/status-uploader.go
schrodit: ok if this is the case then deleting and reupload is better 👍

pkg/testrunner/result/status-uploader.go
OlegLoewen: To use NewUploadRequest you need the repository client service, which is private unfortunately. Seems like it's not possible to use NewUploadRequest like you suggested. What do you think?

schrodit: Its private for the repositories struct but you have direct access to the client as it is the default `github.Client`.
So just use `client.NewUploadReques()` instead of `client.Repositories.NewUploadRequest `

OlegLoewen:
added additional features:
- upload assets to draft release
- asset name prefix
- upload only for successful testruns
- multi component support

please rereview

cmd/testrunner/cmd/run_template/run_template.go
```diff
@@ -44,6 +44,8 @@ var (
 
 	timeout  int64
 	interval int64
+
+	componentsListString string
```
schrodit: where do we need this var?

pkg/testrunner/result/collect.go
```diff
@@ -78,6 +70,61 @@ func (c *Collector) Collect(log logr.Logger, tmClient kubernetes.Interface, name
...
+			err := MarkTestrunsAsIngested(runLogger, tmClient, run.Testrun)
+			if err != nil {
+				runLogger.Error(err, "unable to ingest testrun")
+			}
+		}
+	}
+}
+
+func (c *Collector) uploadStatusAsset(cfg Config, runLogger logr.Logger, err error, run *testrunner.Run, tmClient kubernetes.Interface) {
+	// upload asset only if testrun was successful
```
schrodit: comment can be removed

PR #157, opened 2019-10-01
Enhance authorization

schrodit:
Enhances the authorization mechanism of the github bot.
It is now possible for every plugin to define its own authorization level
- `all`: everyone is allowed to call the plugin
- `org`: only org members are allowed to call the plugin
- `team`: only the requested team is allowed to call the plugin

org admins are always allowed.

**Release note**:

```improvement operator
Enhanced the authorization of the github bot
```

### DIFF
--- a/pkg/tm-bot/github/client.go
+++ b/pkg/tm-bot/github/client.go
@@ -51,14 +51,14 @@ func (c *client) ResolveConfigValue(event *GenericRequestEvent, value *ghval.Git
 		return *value.Value, nil
 	}
 	if value.PRHead != nil && *value.PRHead {
-		pr, _, err := c.client.PullRequests.Get(context.TODO(), event.GetOwnerName(), event.GetRepositoryName(), event.Number)
+		pr, err := c.getPullRequest(event)
 		if err != nil {
 			return "", pluginerr.New("unable to get pr", err.Error())
 		}
 		return pr.GetHead().GetSHA(), nil
 	}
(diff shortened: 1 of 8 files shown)


dguendisch:
> org admins are always allowed

Why? Any reason for that special treatment? Normally you'd always allow admins only to avoid lockouts. However I don't see any lockout possibility here.

pkg/tm-bot/github/client.go
```diff
@@ -123,18 +123,79 @@ func (c *client) UpdateStatus(event *GenericRequestEvent, state State, statusCon
...
+	if MembershipStatus(membership.GetState()) != MembershipStatusActive {
+		return false
+	}
+	if MembershipRole(membership.GetRole()) == MembershipRoleAdmin {
+		return true
+	}
+	return false
+}
+
+// isInRequestedTeam checks if the author is in the organization
```
dguendisch:
```suggestion
// isInOrganization checks if the author is in the organization
```

schrodit:
> > org admins are always allowed
> 
> Why? Any reason for that special treatment? Normally you'd always allow admins only to avoid lockouts. However I don't see any lockout possibility here.

Exactly but currently there could be a lockout if no team is defined only requested reviewers.
This could be the case for some repos that only have a codeowners file with only hardcoded members

schrodit:
closed in favor of https://github.com/gardener/test-infra/pull/158

PR #168, opened 2019-10-17
Refactor and fix e2e junit xml parsing and test summary calculation

OlegLoewen:
Refactor and fix e2e junit xml parsing and test summary calculation

**Release note**:

```
Refactor and fix e2e junit xml parsing and test summary calculation
```

### DIFF
--- a/integration-tests/e2e/kubetest/results_evaluator.go
+++ b/integration-tests/e2e/kubetest/results_evaluator.go
@@ -66,70 +66,111 @@ func writeSummaryToFile(summary Summary) {
 }
 
 func analyzeJunitXMLsEnrichSummary(junitXMLFilePaths []string, summary *Summary) error {
-	var mergedJunitXmlResult = &JunitXMLResult{FailedTests: 0, ExecutedTests: 0, DurationFloat: 0, SuccessfulTests: 0, DurationInt: summary.TestsuiteDuration}
-	testcaseNameToTestcase := make(map[string]TestcaseResult)
-	failureOccurrences := make(map[string]int)
+	var mergedJunitXmlResult = &JunitXMLResult{DurationInt: summary.TestsuiteDuration}
+	var testcases []TestcaseResult
+	failureOccurrences := make(map[string]int) // map of testcases that failed at least once
+	succeededTestcases := make(map[string]bool) // map of testcases that succeeded at east once
+	skippedTestcases := make(map[string]TestcaseResult)
+
 	for _, junitXMLPath := range junitXMLFilePaths {
-		file, err := os.Open(junitXMLPath)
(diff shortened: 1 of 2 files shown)


integration-tests/e2e/kubetest/results_evaluator.go
```diff
@@ -66,70 +66,111 @@ func writeSummaryToFile(summary Summary) {
 }
 
 func analyzeJunitXMLsEnrichSummary(junitXMLFilePaths []string, summary *Summary) error {
-	var mergedJunitXmlResult = &JunitXMLResult{FailedTests: 0, ExecutedTests: 0, DurationFloat: 0, SuccessfulTests: 0, DurationInt: summary.TestsuiteDuration}
-	testcaseNameToTestcase := make(map[string]TestcaseResult)
-	failureOccurrences := make(map[string]int)
+	var mergedJunitXmlResult = &JunitXMLResult{DurationInt: summary.TestsuiteDuration}
```
schrodit:
```suggestion
	mergedJunitXmlResult := &JunitXMLResult{DurationInt: summary.TestsuiteDuration}
```
nit/ just to be in sync with the assignments in 71-74

integration-tests/e2e/kubetest/results_evaluator.go
```diff
@@ -66,70 +66,111 @@ func writeSummaryToFile(summary Summary) {
...
+			// write only not skipped testcases to json file
+			if err := writeTestcaseToJSONFile(newTestcase); err != nil {
+				return err
+			}
+		}
+	}
+	return nil
+}
+
+func unmarshalJUnitXMLResult(junitXMLPath string) (JunitXMLResult, error) {
```
schrodit: maybe rename to `unmashalJUnitFromFile` because there is already a method called `UnmarshalJunitXMLResult`. Just that the difference is clear

integration-tests/e2e/main.go
```diff
@@ -45,7 +46,10 @@ func main() {
 
 func createDescFileOfFailedTestcases(failedTestcases []string) string {
 	generatedRunDescPath := filepath.Join(config.TmpDir, "failedTestcasesDescription.txt")
-	file, err := os.OpenFile(generatedRunDescPath, os.O_APPEND|os.O_CREATE|os.O_WRONLY, 0644)
+	if err := os.Remove(kubetest.GeneratedRunDescPath); err != nil {
+		log.Error(err)
```
schrodit: shouldn't we exit if we are unable to remove the file?
not just log the error or is the deletion just a nice to have cleanup?

OlegLoewen: this is only necessary, if e2e.go is executed multiple times, which happens only via local testing.

PR #182, opened 2019-11-08
Add basic ui for tm bot

schrodit:
Adds a basic ui with documentation for the testmachinery bot so that w can rely on the generic documentation and do not have to manually maintain the markdowns.

In the future is is also planned to have also pages for running tests and the overall status of the bot and the testmachinery (maybe also direct links to grafana and argo).

image

image

**Release note**:

```improvement operator
Add basic implementation of Test Machinery bot UI.
```


pkg/tm-bot/plugins/plugins.go
```diff
@@ -116,6 +121,15 @@ func (p *Plugins) Get(name string) (string, Plugin, error) {
 	return runID, plugin.New(runID), nil
 }
 
+// GetAll returns all registred plugins
```
OlegLoewen:
```suggestion
// GetAll returns all registered plugins
```

pkg/tm-bot/router.go
```diff
@@ -45,10 +46,11 @@ func setup(log logr.Logger) (*mux.Router, error) {
 
 	r := mux.NewRouter()
 	r.Use(loggingMiddleware(log.WithName("trace")))
-
 	r.HandleFunc("/healthz", healthz(log.WithName("health"))).Methods(http.MethodGet)
 	r.HandleFunc("/events", hooks.HandleWebhook).Methods(http.MethodPost)
 
+	botUI := ui.New(log, "/Users/d064999/go/src/github.com/gardener/test-infra/pkg/tm-bot/ui")
```
dguendisch: That path looks wrong :)

PR #195, opened 2019-11-25
Introduce flake attempts for TM testruns

OlegLoewen:
Introduce flake attempts for TM testruns

**Release note**:

```
Introduce flake attempts for TM testruns
```

### DIFF
--- a/pkg/apis/testmachinery/v1beta1/types.go
+++ b/pkg/apis/testmachinery/v1beta1/types.go
@@ -135,6 +135,9 @@ type TestrunStatus struct {
 	// This value is calculated by (CompletionTime - StartTime)
 	Duration int64 `json:"duration,omitempty"`
 
+	// Tells whether testrun failed first and succeeded in a retry
+	Flaked bool `json:"flaked,omitempty"`
+
 	// Workflow is the name of the generated argo workflow
 	Workflow string `json:"workflow,omitempty"`
 
--- a/pkg/apis/testmachinery/v1beta1/types.go
+++ b/pkg/apis/testmachinery/v1beta1/types.go
@@ -142,7 +142,7 @@ type TestrunStatus struct {
 	// It also shows all specific executed tests.
 	Steps []*StepStatus `json:"steps,omitempty"`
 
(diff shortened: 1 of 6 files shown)


pkg/apis/testmachinery/v1beta1/types.go
```diff
@@ -135,6 +135,9 @@ type TestrunStatus struct {
 	// This value is calculated by (CompletionTime - StartTime)
 	Duration int64 `json:"duration,omitempty"`
 
+	// Tells whether testrun failed first and succeeded in a retry
+	Flaked bool `json:"flaked,omitempty"`
```
schrodit: do not make it part of the testrun resource as this parameter is only known by the testrunner

maybe rather add it to the Run or directly to the metadata

pkg/testrunner/types.go
```diff
@@ -117,6 +120,7 @@ type TestrunSummary struct {
 	Phase     argov1.NodePhase `json:"phase,omitempty"`
 	StartTime *metav1.Time     `json:"startTime,omitempty"`
 	Duration  int64            `json:"duration,omitempty"`
+	Flaked    bool             `json:"flaked,omitempty"`
```
schrodit: remove see comment above

pkg/testrunner/testrunner.go
```diff
@@ -54,19 +54,35 @@ func (rl RunList) Run(log logr.Logger, tmClient kubernetes.Interface, namespace,
...
+				}
+				rl[i].Error = err
+
+				if err == nil && tr.Status.Phase == tmv1beta1.PhaseStatusSuccess {
+					if flakeAttempt != 0 {
+						tr.Status.Flaked = true
+					}
+					// testrun was successful, break retry loop
+					break
+				} else {
```
schrodit: else in not necessary as we break out in the if branch anyways

OlegLoewen: if flakeAttempts > 1, the else branch ist required

pkg/apis/testmachinery/v1beta1/types.go
OlegLoewen: done

PR #196, opened 2019-11-25
Add a few pre computed fields to teststeps (for better grafana int)

dguendisch:
Adds a few pre computed fields to teststeps for better grafana ingegration.

Fixes #

**Release note**:

```improvement operator

```

### DIFF
--- a/pkg/testrunner/result/output_test.go
+++ b/pkg/testrunner/result/output_test.go
@@ -2,12 +2,12 @@ package result_test
 
 import (
 	"github.com/gardener/test-infra/pkg/apis/testmachinery/v1beta1"
+	"github.com/gardener/test-infra/pkg/logger"
 	"github.com/gardener/test-infra/pkg/testrunner"
 	"github.com/gardener/test-infra/pkg/testrunner/result"
 	"github.com/gardener/test-infra/test/resources"
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
-
 	"testing"
 )
 
@@ -17,7 +17,7 @@ func TestValidationWebhook(t *testing.T) {
 }
 
 var _ = Describe("output generation tests", func() {
-
+	log := logger.Log.WithName("output_test")
 	Context("configuration", func() {
 		It("should include environment variable configuration to in the metadata", func() {
 			configElement := v1beta1.ConfigElement{
@@ -39,7 +39,7 @@ var _ = Describe("output generation tests", func() {
 				},
 			}
 
-			_, summaries, err := result.DetermineTestrunSummary(tr, &testrunner.Metadata{}, &result.Config{})
+			_, summaries, err := result.DetermineTestrunSummary(tr, &testrunner.Metadata{}, &result.Config{}, nil, log)
 			Expect(err).ToNot(HaveOccurred())
 
 			Expect(summaries).To(HaveLen(1))
@@ -68,8 +68,7 @@ var _ = Describe("output generation tests", func() {
 				},
 			},
 		}
-
-		_, summaries, err := result.DetermineTestrunSummary(tr, &testrunner.Metadata{}, &result.Config{})
+		_, summaries, err := result.DetermineTestrunSummary(tr, &testrunner.Metadata{}, &result.Config{}, nil, log)
 		Expect(err).ToNot(HaveOccurred())
 
 		Expect(summaries).To(HaveLen(1))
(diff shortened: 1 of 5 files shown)


pkg/testrunner/run.go
```diff
@@ -137,3 +139,23 @@ func GetArgoURL(tmClient kubernetes.Interface, tr *tmv1beta1.Testrun) (string, e
 
 	return argoUrl.String(), nil
 }
+
+// GetClusterDomainURL tries to derive the cluster domain url from an grafana ingress if possible. Returns an error if the ingress cannot be found or is in unexpected form.
+func GeClusterDomainURL(tmClient kubernetes.Interface) (string, error) {
```
OlegLoewen:
```suggestion
func GetClusterDomainURL(tmClient kubernetes.Interface) (string, error) {
```

pkg/testrunner/result/output_test.go
```diff
@@ -17,7 +17,7 @@ func TestValidationWebhook(t *testing.T) {
 }
 
 var _ = Describe("output generation tests", func() {
-
+	log := logger.Log.WithName("output_test")
```
schrodit: lets use the `NullLogger` if we do not need any log output in the unit tests?

pkg/testrunner/run.go
```diff
@@ -23,8 +23,10 @@ import (
 	"github.com/go-logr/logr"
 	"github.com/hashicorp/go-multierror"
 	"k8s.io/api/extensions/v1beta1"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
```
schrodit: nit
```suggestion
	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
```

pkg/testrunner/run.go
```diff
@@ -137,3 +139,23 @@ func GetArgoURL(tmClient kubernetes.Interface, tr *tmv1beta1.Testrun) (string, e
...
+	// return err if the ingress cannot be found
+	ingress, err := tmClient.Kubernetes().ExtensionsV1beta1().Ingresses("monitoring").Get("grafana", v1.GetOptions{})
+	if err != nil {
+		return "", fmt.Errorf("cannot get grafana ingress: %v", err)
+	}
+	if len(ingress.Spec.Rules) == 0 {
+		return "", fmt.Errorf("cannot get ingress rule from ingress %v", ingress)
+	}
+	host := ingress.Spec.Rules[0].Host
+	r, _ := regexp.Compile("[a-z]+\\.ingress\\.(.+)$")
```
schrodit: please handle the error

pkg/testrunner/types.go
```diff
@@ -122,11 +122,26 @@ type TestrunSummary struct {
...
+	Phase       argov1.NodePhase `json:"phase,omitempty"`
+	StartTime   *metav1.Time     `json:"startTime,omitempty"`
+	Duration    int64            `json:"duration,omitempty"`
+	PreComputed *StepPreComputed `json:"pre,omitempty"`
+}
+
+// StepPreComputed contains fields that could be created at runtime via scripted fields, but are created statically for better performance and better support of grafana
+type StepPreComputed struct {
+	// same as StepSummary.Phase but mapping states to ints (Failed&Timeout -> 0, Succeeded -> 100); allows to do averages on success rate in dashboards
+	PhaseNum int `json:"phaseNum,omitempty"`
```
schrodit: We should change this to a pointer because otherwise all testruns that are `phase != sucess` will be 0 as the struct will be automatically initialized with a value of 0.
If we change the type to a pointer there will be a difference between 0 and nil.

```suggestion
	PhaseNum *int `json:"phaseNum,omitempty"`
```

pkg/testrunner/result/output_test.go
dguendisch: sure, done

pkg/testrunner/run.go
dguendisch: oops :)

pkg/testrunner/run.go
dguendisch: done

pkg/testrunner/run.go
dguendisch: I intentionally did not: the regex is static -> cannot change during runtime -> cannot error during runtime

pkg/testrunner/types.go
dguendisch: Good catch! Thanks!

pkg/testrunner/run.go
schrodit: oh yes you're right 👍

PR #198, opened 2019-11-26
Refactor testrunner and add authentication for the github bot

schrodit:
The testrunner now uses watches (uses a controller runtime controller) to watch testruns.
This should also decrease the number of logs

**Release note**:

```improvement operator
Refactor testrunner to use watches on testruns instead of polling
```
```improvement operator
Add oauth authentication to the testrunner to protect several pages and internal information
```


pkg/tm-bot/ui/auth/auth.go
```diff
@@ -0,0 +1,196 @@
...
+		client := github.NewClient(tc)
+
+		membership, _, err := client.Organizations.GetOrgMembership(ctx, aCtx.User, a.org)
+		if err != nil {
+			a.log.Error(err, "unable to get org membership")
+			http.Redirect(w, r, "/404", http.StatusTemporaryRedirect)
+			return
+		}
+		if github2.MembershipStatus(membership.GetState()) != github2.MembershipStatusActive {
+			a.log.Error(err, "user member of org", "user", aCtx.User, "org", a.org)
```
dguendisch:
```suggestion
			a.log.Error(err, "user not active member of org", "user", aCtx.User, "org", a.org)
```

PR #206, opened 2019-12-10
Add resource requests/limits to argo workflow controller deployment

dguendisch:
Added configurable resource requests/limits to argo workflow controller deployment

**Release note**:
```improvement operator
Argo workflow controller deployment has now configurable resource requests/limits
```


charts/bootstrap_tm_prerequisites/templates/argo.yaml
```diff
@@ -116,4 +116,6 @@ spec:
         - workflow-controller
         image: argoproj/workflow-controller:{{ .Values.argo.tag }}
         name: workflow-controller
+        resouces:
```
schrodit:
```suggestion
        resources:
```

dguendisch: oops

PR #208, opened 2019-12-10
Added Favicon

dguendisch:
What's a website without a favicon :)
PS: didn't want to add bitmaps, so re-using the svg one => this works today in Firefox, Chrome will support it with v80 (Feb 2020)


dguendisch:
cherry-picked and rebased :)

PR #231, opened 2020-02-04
Refactor status of ingested and uploaded fields

schrodit:

The `ingestion` and `github upload` state are removed from the status of the testrun.
These tes are replaced by 2 labels:
- `testrunner.testmachinery.sapcloud.io/ingested`
- `testrunner.testmachinery.sapcloud.io/uploaded-to-github`

**Release note**:

```improvement operator
⚠️ The `ingested` and `uploadedToGithHub` attributes in the state are removed and replaced by the labels `testrunner.testmachinery.sapcloud.io/ingested` and `testrunner.testmachinery.sapcloud.io/uploaded-to-github`
```

### DIFF
--- a/pkg/util/util.go
+++ b/pkg/util/util.go
@@ -24,6 +24,7 @@ import (
 	"github.com/pkg/errors"
 	"io"
 	"io/ioutil"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"math"
 	"math/rand"
 	"net/http"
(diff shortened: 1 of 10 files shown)


pkg/util/util.go
```diff
@@ -218,27 +219,42 @@ func StringDefault(value, def string) string {
 	return value
 }
 
+// HasAnnotation returns a bool if passed in label exists
```
OlegLoewen:
```suggestion
// HasAnnotation returns a bool if passed in label exists
```
```suggestion
// HasLabel returns a bool if passed in label exists
```

PR #233, opened 2020-02-04
add new run to execution queue

schrodit:
Fixes the retries with the new executor by re-adding the current function to the queue

**Release note**:

```improvement operator
Fix retries with new executor
```

### DIFF
--- a/pkg/testrunner/list.go
+++ b/pkg/testrunner/list.go
@@ -78,9 +78,9 @@ func (rl RunList) Run(log logr.Logger, config *Config, testrunNamePrefix string,
 		var (
 			trI     = i
 			attempt = 0
+			f       func()
 		)
-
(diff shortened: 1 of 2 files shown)


pkg/testrunner/list.go
```diff
@@ -108,7 +108,9 @@ func (rl RunList) Run(log logr.Logger, config *Config, testrunNamePrefix string,
 			}
 			*rl[trI] = *newRun
 			attempt++
-		})
+			executor.AddItem(f)
+		}
+		executor.AddItem(f)
```
OlegLoewen: Can you explain why you need to add f twice? Looks confusing

schrodit: I add `f` once for the initial run and then for the retries inside the execution

PR #243, opened 2020-02-12
prevent redundant elasticsearch ingestions

OlegLoewen:
prevent redundant elasticsearch ingestions

**Release note**:

```
NONE
```

### DIFF
--- a/pkg/util/util.go
+++ b/pkg/util/util.go
@@ -404,7 +404,7 @@ func Zipit(source, target string) error {
 
 // DocExists checks whether an elasticsearch doc of a testrun exists in index testmachinery-* index
 func DocExists(log logr.Logger, esConfig *elasticsearch.Config, testrunID, testrunStartTime string) (docExists bool) {
-	log.Info("check if docs have already been ingested")
+	log.V(2).Info("check if docs have already been ingested")
 
 	payload := fmt.Sprintf(`{
 			"size": 0,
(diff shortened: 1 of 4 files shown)


pkg/util/util.go
```diff
@@ -399,3 +401,48 @@ func Zipit(source, target string) error {
 
 	return err
 }
+
+// DocExists checks whether an elasticsearch doc of a testrun exists in index testmachinery-* index
+func DocExists(log logr.Logger, esConfig *elasticsearch.Config, testrunID, testrunStartTime string) (docExists bool) {
+	log.Info("check if docs have already been ingested")
```
schrodit:
```suggestion
	log.V(2).Info("check if docs have already been ingested")
```

I think the info not needed in every run

pkg/util/util.go
```diff
@@ -399,3 +401,48 @@ func Zipit(source, target string) error {
...
+
+	if esConfig == nil {
+		log.Info("no elasticsearch config available, cannot check if doc already exists")
+		return false
+	}
+	esClient, err := elasticsearch.NewClient(*esConfig)
+	if err != nil {
+		log.Error(err, "couldn't create elasticsearch client")
+		return false
+	}
```
schrodit: we already create a client in the `ingestDir`, maybe we should only pass the client to have the same code twice.

OlegLoewen: this makes sense after removal of deprecated es.configname approach. Therefore postponed

PR #249, opened 2020-02-24
Fix busyloop blocking goroutine

schrodit:

Fixes #

**Release note**:

```improvement operator
Fixed the problem that testrun watches are blocked in serial executed testruns
```


pkg/util/waitgroup.go
```diff
@@ -45,11 +46,9 @@ func (wg *AdvancedWaitGroup) Done() {
 
 // Wait waits until the counter is 0
 func (wg *AdvancedWaitGroup) Wait() {
-	for {
-		if wg.count == 0 {
-			return
-		}
-	}
+	wg.WaitWithCancelFunc(func() bool {
+		return true
```
dguendisch:
```suggestion
		return false
```
:)

PR #252, opened 2020-03-02
move slack testruns summary posting to collect package

OlegLoewen:
move slack testruns summary posting to collect package

**Release note**:

```
Moving testruns summary posting to collect package
```


pkg/testrunner/result/summary-poster.go
```diff
@@ -0,0 +1,200 @@
...
+		log.Error(err, "failed creating a table to post")
+	}
+	if table == "" {
+		log.Info("no table to render")
+		return
+	}
+
+	slackClient, err := slack.New(log, config.SlackToken)
+	if err != nil {
+		log.Error(err, "Was not able to create slack client")
```
schrodit:
```suggestion
		log.Error(err, "was not able to create slack client")
```

pkg/testrunner/result/summary-poster.go
```diff
@@ -0,0 +1,200 @@
...
+	}
+
+	table.SetHeader(header)
+	table.AppendBulk(res.GetContent())
+	table.SetAlignment(tablewriter.ALIGN_CENTER)
+	table.Render()
+	return writer.String(), nil
+}
+
+type resultRow struct {
```
schrodit: looks like all the sorting and maybe some parts of the rendering could be abtracted and shared with the functions from https://github.com/gardener/test-infra/blob/9aad6bdac94870dff02aefa692881d777a02b550/cmd/testrunner/cmd/notify/tablecontent.go#L29

I think there could be a common renderner in `utils` that takes an abstracted model to generate the table something like an inner join of `testrunner.Run` and `result.AssetOverviewItems`? wdyt?

OlegLoewen: can you check again?

PR #255, opened 2020-03-10
Add bot plugin to run simple testruns

schrodit:
A new plugin has been added to execute a single testrun

I'm unsure about the naming scheme also regarding new test that may be added in the future.
The current naming scheme is `test-[name]`. Any better ideas?

**Release note**:

```improvement user
A new bot plugin has been added that the execution of simple testruns that reside in the repository.
A single testrun can be execution by typing `/test-single [path to the testrun]`.
```
```action user
The old `test` command has been renamed to `test-gardener` which means that gardener test can now only be invoked via `/test-gardener`
```


pkg/common/error/error.go
```diff
@@ -0,0 +1,124 @@
...
+func NewTimeoutError(message string) error {
+	return newError(ReasonTimeout, message)
+}
+
+// NewNotCreatedError returns an error indicating that a resource could not be created
+func NewNotCreatedError(message string) error {
+	return newError(ReasonNotCreated, message)
+}
+
+// NewNotCreatedError returns an error indicating that a resource could not be created
```
OlegLoewen:
```suggestion
// NewNotRenderedError returns an error indicating that a resource could not be created
```

pkg/tm-bot/github/authorization.go
```diff
@@ -0,0 +1,118 @@
...
+			return false
+		}
+		if MembershipStatus(membership.GetState()) == MembershipStatusActive {
+			return true
+		}
+	}
+	return false
+}
+
+// isInRequestedTeam checks if the author is in the requested PR team
```
OlegLoewen:
```suggestion
// isInDefaultTeam checks if the author is in the requested PR team
```

PR #265, opened 2020-04-08
Refactor prerequisites to use resource manager

### LINKED ISSUE #263: Deploy prerequisiteis at runtime
**What would you like to be added**:

The prerequisites chart (containing argo workflows, crds, minio, ..) should be automatically installed during runtime so it can be also automatically updated and checked if all needed components are available.

Therefore we should have a look at https://github.com/gobuffalo/packr to include the complete helmchart.
We could also think of reusing the gardener resource manager(https://github.com/gardener/gardener-resource-manager) to ensure everything is up and running and gets reconciled otherwise we would have to manage leaked resource on our own.

as a result I want to move all configuration to a dedicated configuration file that then also includes the helm values for the additional deployed charts.

WDYT @dguendisch @OlegLoewen

schrodit:

Fixes #263

Still WIP:
- [x] adapt local dev setup
- [x] adapt docs

**Release note**:

```improvement operator
The TestMachinery Configuration is reworked so that now configuration is done via a configuration file.
The path to the configuration can be specified via `--config` flag.
⚠️ most flags were removed and replaced by the configuration file.
```
```improvement operator
The TestMachinery prerequisites deployment was refactored to now be deployed reconciled by the TestMachinery itself.
The prerequisites chart was removed and is not needed anymore.
```


Makefile
```diff
@@ -73,17 +85,19 @@ gen-certs:
...
-	@ if [ ${NS} != "default" ] && [ kubectl get ns ${NS} ]; then \
-		kubectl delete ns ${NS}; \
-	fi
+################################
+# Prerequisistes, Installation #
+################################
+
+.PHONY: deploy-controller
+deploy-controller:
+	helm template --namespace ${NS} -f ./charts/testmachinery/local-values.yaml --set "controller.tag=${VERSION}" ./charts/testmachinery | kubectl create -f -
```
dguendisch, line 95: Doesn't `kubectl apply` make more sense, as in e.g. deploying the controller subsequently after changing code?

charts/testmachinery/charts/gardener-resource-manager/templates/deployment.yaml
```diff
@@ -0,0 +1,64 @@
...
+# limitations under the License.
+
+---
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: gardener-resource-manager
+  namespace: {{ .Release.Namespace }}
+  labels:
+    garden.sapcloud.io/role: controlplane
```
dguendisch, line 22: Do you think such labels could interfere with the "real" resource manager?

charts/testmachinery/charts/gardener-resource-manager/values.yaml
```diff
@@ -0,0 +1,22 @@
+images:
+  gardener-resource-manager: eu.gcr.io/gardener-project/gardener/gardener-resource-manager:v0.12.0
+
+resources:
+  requests:
+    cpu: 23m
+    memory: 47Mi
```
dguendisch, line 7: That looks like magic numbers, but...OK :)

charts/testmachinery/templates/crd.yaml
```diff
@@ -0,0 +1,53 @@
...
+    type: string
+    description: The phase indicates the current status of the overall testrun.
+    JSONPath: .status.phase
+  - name: StartTime
+    type: date
+    description: The StarTime indicates the time the testrun was triggered.
+    JSONPath: .status.startTime
+  - name: Duration
+    type: number
+    description: The Duration inidcates the complete duration of the workflow.
```
dguendisch, line 52:
```suggestion
    description: The Duration indicates the complete duration of the workflow.
```

pkg/apis/config/config.go
```diff
@@ -0,0 +1,167 @@
...
+	// Namespace is the namespace the testmachinery is deployed to.
+	Namespace string `json:"namespace,omitempty"`
+
+	// TestDefPath is the repository path where the Test Machinery should search for testdefinitions.
+	TestDefPath string `json:"testdefPath"`
+
+	// PrepareImage is the prepare image that is used in the prepare and postprepare step.
+	PrepareImage string `json:"prepareImage"`
+
+	// PrepareImage is the base image that is used as the default image if a TestDefinition does not define a image.
```
dguendisch, line 76:
```suggestion
	// PrepareImage is the base image that is used as the default image if a TestDefinition does not define an image.
```

pkg/apis/config/v1beta1/config.go
```diff
@@ -0,0 +1,167 @@
...
+	// Namespace is the namespace the testmachinery is deployed to.
+	Namespace string `json:"namespace,omitempty"`
+
+	// TestDefPath is the repository path where the Test Machinery should search for testdefinitions.
+	TestDefPath string `json:"testdefPath"`
+
+	// PrepareImage is the prepare image that is used in the prepare and postprepare step.
+	PrepareImage string `json:"prepareImage"`
+
+	// PrepareImage is the base image that is used as the default image if a TestDefinition does not define a image.
```
dguendisch, line 76:
```suggestion
	// PrepareImage is the base image that is used as the default image if a TestDefinition does not define an image.
```

charts/testmachinery/charts/gardener-resource-manager/templates/deployment.yaml
schrodit: yes you're right.
Forgot to remove these labels during copying.

I just removed all of them

charts/testmachinery/charts/gardener-resource-manager/values.yaml
schrodit: just copied values. but I gues its ok to make them not look htat magically:)

schrodit:
> maybe you could check your copyright template (most new files point to 2019, a few files point to 2020 :) )

My template is correct. the problem was that I copied some files.
I think I will open a separate PR that updates all Licenses.

PR #269, opened 2020-04-24
Update/argo

### LINKED ISSUE #262: add direct grafana/loki links to argo workflow resource
Argo 2.7.0 introduces the possibility to directly link to external logging facitilies.
Add the respective grafana/loki links to the argo resources (workflow).
See https://github.com/argoproj/argo/commit/14478bc07f42ae9ee362cc1531b1cf00d923211d

schrodit:

image

Fixes #262

currently I only added links to to grafana/loki, maybe it would make sense to also add a workflow link to the TM Dashboard. WDYT?
If we wanted to do this we would have to refactor the workflow naming from `<testrun>-wf` to `<testrun>`. But I think this would then also be more convinient.

**Release note**:

```improvement operator
Update argo to v2.7.5
```
```improvement operator
Add new argo ui supported log links to grafana/loki logging
```


Makefile
```diff
@@ -43,6 +43,7 @@ LD_FLAGS := $(shell ./hack/get-build-ld-flags)
 revendor:
 	@GO111MODULE=on go mod vendor
 	@GO111MODULE=on go mod tidy
+<<<<<<< HEAD
```
dguendisch, line 46: Merging/Rebasing probably went wrong :)

Makefile
```diff
@@ -56,6 +57,21 @@ mock-gen:
...
+
+.PHONY: mock-gen
+mock-gen:
+	@./hack/generate-mocks
+
+.PHONY: generate
+generate:
+	@$(REPO_ROOT)/vendor/github.com/gardener/gardener-extensions/hack/generate.sh ./...
+
+>>>>>>> a13a36b7... Refactor prerequisites to use resource manager
```
dguendisch, line 74: Merging/Rebasing probably went wrong :)

pkg/testrunner/util.go
```diff
@@ -42,23 +42,28 @@ func GetArgoURLFromHost(host string, tr *tmv1beta1.Testrun) string {
...
-	return GetHostURLFromIngress(tmClient, client.ObjectKey{Name: "argo-ui", Namespace: "default"})
+func GetArgoHost(ctx context.Context, tmClient client.Client) (string, error) {
+	return GetHostURLFromIngress(ctx, tmClient, client.ObjectKey{Name: "argo-ui", Namespace: "default"})
 }
 
 // GetGrafanaURLFromHostForWorkflow returns the path to the logs in grafana for a whole workflow
-func GetGrafanaURLFromHostForWorkflow(host string, tr *tmv1beta1.Testrun) string {
-	return fmt.Sprintf(`%s/explore?left=["now-7d","now","Loki",{"expr":"{container%%3D\"main\",argo_workflow%%3D\"%s\"}"},{"mode":"Logs"},{"ui":[true,true,true,"exact"]}]`, host, tr.Status.Workflow)
+func GetGrafanaURLFromHostForWorkflow(host string, workflowName string) string {
+	return fmt.Sprintf(`%s/explore?left=["now-7d","now","Loki",{"expr":"{container%%3D\"main\",argo_workflow%%3D\"%s\"}"},{"mode":"Logs"},{"ui":[true,true,true,"exact"]}]`, host, workflowName)
```
dguendisch, line 51: Maybe make a constant out of `7d` and set it to `3d` as workflows older >2d shouldn't be present in the system anyway.

dguendisch:
A link to the TM Dashboard would be great as well but is not that desperately needed probably (as most often I navigate from the TM Dashboard to Argo and not vice versa). At some point in time we need to check the concourse URL, that is sometimes important to understand what ultimately triggered the respective workflow.

pkg/testrunner/util.go
schrodit: thats true for most of our pods.
But I think 3d is totally ok.
For the tm dashboard we could also think if directly enter the start and end time of the needed resource. But just a future improvement

Issue #279, opened 2020-05-08
Reserve Excess capacity pods are monitored in the TM health check


schrodit:
**What happened**:
The reserve-excess-capacity deployment is monitored via the argo mr.
If the cluster currently scales up the health check fails which causes the tm to be unhealthy

**What you expected to happen**:
The access capacity pods are not considered for healthiness

PR #280, opened 2020-05-08
Use own managed resource for excess capacity pods

schrodit:

Fixes https://github.com/gardener/test-infra/issues/279

**Release note**:

```improvement operator
The reserve-excess-capacity pods are now refactored to use their own ManagedResources and will not monitored by the healthcheck.
```


charts/images.yaml
```diff
@@ -11,6 +11,10 @@ images:
   sourceRepository: github.com/argoproj/argo
   repository: argoproj/argoexec
   tag: "v2.7.5"
+- name: reserve-excess-capacity
+  sourceRepository: github.com/argoproj/argo
```
dguendisch, line 15: Wondering where we had this from but doesn't matter too much :)

PR #287, opened 2020-05-28
add allowPrivilegedContainers to flavors

dguendisch:
Adds `allowPrivilegedContainers` as an additional option in the flavor configuration. Gardener shoots default to `true` is this option is not set, hence you can use it to disable privileged containers for certain tests or have your tests run against both: shoots with `allowPrivilegedContainers="true"` and shoots with `allowPrivilegedContainers="false"`

**Release note**:
```improvement user
`allowPrivilegedContainers` can now be customized in the flavor configuration.
```

### DIFF
--- a/pkg/common/types_shootflavor.go
+++ b/pkg/common/types_shootflavor.go
@@ -47,6 +47,9 @@ type Shoot struct {
 	// Kubernetes versions to test
 	KubernetesVersion gardencorev1beta1.ExpirableVersion
 
+	// AllowPrivilegedContainers defines whether privileged containers will be allowed in the given shoot or not
+	AllowPrivilegedContainers string
+
 	// Worker pools to test
 	Workers []gardencorev1beta1.Worker
 }
@@ -62,6 +65,10 @@ type ShootFlavor struct {
 
 	// Kubernetes versions to test
 	KubernetesVersions ShootKubernetesVersionFlavor `json:"kubernetes"`
+
+	// AllowPrivilegedContainers defines whether privileged containers will be allowed in the given shoot or not
+	// +optional
+	AllowPrivilegedContainers string `json:"allowPrivilegedContainers"`
 
 	// Worker pools to test
 	Workers []ShootWorkerFlavor `json:"workers"`
--- a/pkg/common/types_shootflavor.go
+++ b/pkg/common/types_shootflavor.go
@@ -48,7 +48,7 @@ type Shoot struct {
 	KubernetesVersion gardencorev1beta1.ExpirableVersion
 
 	// AllowPrivilegedContainers defines whether privileged containers will be allowed in the given shoot or not
-	AllowPrivilegedContainers string
+	AllowPrivilegedContainers *bool
 
 	// Worker pools to test
 	Workers []gardencorev1beta1.Worker
@@ -68,7 +68,7 @@ type ShootFlavor struct {
 
 	// AllowPrivilegedContainers defines whether privileged containers will be allowed in the given shoot or not
 	// +optional
-	AllowPrivilegedContainers string `json:"allowPrivilegedContainers"`
+	AllowPrivilegedContainers *bool `json:"allowPrivilegedContainers"`
 
 	// Worker pools to test
 	Workers []ShootWorkerFlavor `json:"workers"`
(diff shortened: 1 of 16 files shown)


pkg/common/types_shootflavor.go
```diff
@@ -63,6 +66,10 @@ type ShootFlavor struct {
 	// Kubernetes versions to test
 	KubernetesVersions ShootKubernetesVersionFlavor `json:"kubernetes"`
 
+	// AllowPrivilegedContainers defines whether privileged containers will be allowed in the given shoot or not
+	// +optional
+	AllowPrivilegedContainers string `json:"allowPrivilegedContainers"`
```
schrodit, line 71: why are you not using a boolean.
if you need to distinguish between tru, false and not set I would suggest to use a boolean pointer?

A string seams strange as you would always require quotes.

dguendisch: Although helm values handling on boolean is a bit annoying (e.g. https://github.com/helm/helm/issues/3308) I switched to bool pointers.

dguendisch:
Regarding metadata, I though it would be enough to distinguish via flavor description, but you are right, a dedicated metadata is better, I'll add that, thanks for the suggestion.

pkg/shootflavors/extendedflavors_test.go
```diff
@@ -82,9 +82,11 @@ var _ = Describe("extended flavor test", func() {
 	})
 
 	It("should return 1 shoot", func() {
+		true := true
 		rawFlavors := []*common.ExtendedShootFlavor{{
 			ExtendedConfiguration: defaultExtendedCfg,
 			ShootFlavor: common.ShootFlavor{
+				AllowPrivilegedContainers: &true,
```
schrodit, line 89: nit: you can just use the k8s pointer package `pointer.BoolPtr(true)`

dguendisch: done

dguendisch:
added metadata support, hope I didn't miss some places, ptal

pkg/testmachinery/metadata/types.go
```diff
@@ -26,9 +26,10 @@ type Metadata struct {
 	Region            string `json:"region,omitempty"`
 
 	// todo: schrodit - add support to better persist multiple worker pools with multiple oss, versions and zones
-	OperatingSystem        string `json:"operating_system,omitempty"`
-	OperatingSystemVersion string `json:"operating_system_version,omitempty"`
-	Zone                   string `json:"zone,omitempty"`
+	OperatingSystem           string `json:"operating_system,omitempty"`
+	OperatingSystemVersion    string `json:"operating_system_version,omitempty"`
+	Zone                      string `json:"zone,omitempty"`
+	AllowPrivilegedContainers *bool  `json:"allowPrivilegedContainers,omitempty"`
```
schrodit, line 32: shouldn't it be snake case `allow_privileged_containers ` as operating system? 
I can't really what we agreed on for the elasticsearch metadata structure

dguendisch: Guess it doesn't matter for elasticsearch, but of course it should follow what we already, so snake case is the way, thanks for spotting.

dguendisch:
> we also need to read the metadta from the testrun here https://github.com/gardener/test-infra/blob/master/pkg/testmachinery/metadata/metadata.go#L57

Added.

> How about adding it also here https://github.com/gardener/test-infra/blob/master/pkg/testmachinery/metadata/metadata.go#L41 so that it will also be displayed in the tm dashboard and the slack notifications?

I actually envisioned to put it in the description of the flavor configuration, but sure, putting it directly in the code is more consistent in the end, so added it.

pkg/testmachinery/metadata/metadata.go
```diff
@@ -72,4 +79,10 @@ func FromTestrun(tr *tmv1beta1.Testrun) *Metadata {
 			ExecutionGroup: tr.Annotations[common.LabelTestrunExecutionGroup],
 		},
 	}
+	if tr.Annotations[common.AnnotationAllowPrivilegedContainers] != "" {
```
schrodit, line 82: nit: I think this is just more convinient. 
Use `metav1.HasAnnotation()` or 
```
if value, ok := tr.Annotations[common.AnnotationAllowPrivilegedContainers]; ok {
 strconv.ParseBool(value)
}

```

dguendisch: went for the second suggestion (for the first I wouldn't have known where to get the correctly typed ObjectMeta from :) )

PR #294, opened 2020-07-02
Adds field shootAnnotations to flavors

dguendisch:
Adds field `annotations` to flavors. It allows to add annotations to created shoot clusters (i.e. to trigger certain shoot behaviour).
Usage example:
```
flavors:
- provider: aws
  annotations:
    a: b
    c: d
```

Cooperates with https://github.com/gardener/gardener/pull/2550

**Release note**:
```improvement user
Annotations for shoot clusters can be customized in a flavor by adding them to field `annotations`.
```

### DIFF
--- a/pkg/common/common.go
+++ b/pkg/common/common.go
@@ -72,6 +72,9 @@ const (
 	// AnnotationAllowPrivilegedContainers is the annotation describing whether and how created shoots will have allowPrivilegedContainers configured
 	AnnotationAllowPrivilegedContainers = "metadata.testmachinery.gardener.cloud/allow-privileged-containers"
 
+	// AnnotationShootAnnotations is the annotation describing which additional shoot annotations were set (as they could impact the shoot behaviour)
+	AnnotationShootAnnotations = "metadata.testmachinery.gardener.cloud/shoot-annotations"
+
 	// AnnotationFlavorDescription is the annotation to describe the test flavor of the current run testrun
 	AnnotationFlavorDescription = "metadata.testmachinery.gardener.cloud/flavor-description"
 
--- a/pkg/testmachinery/metadata/types.go
+++ b/pkg/testmachinery/metadata/types.go
@@ -26,10 +26,11 @@ type Metadata struct {
 	Region            string `json:"region,omitempty"`
 
 	// todo: schrodit - add support to better persist multiple worker pools with multiple oss, versions and zones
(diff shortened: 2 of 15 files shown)


pkg/common/common.go
```diff
@@ -115,3 +123,11 @@ var (
 	// Default timeout of 4 hours to wait before resuming the testrun
 	DefaultPauseTimeout = 14400
 )
+
+func MarshalMap(annotations map[string]string) string {
```
schrodit, line 127: should be better in utils as common currently only contains definitions.

Doc is missing

pkg/testrun_renderer/templates/shoots.go
```diff
@@ -30,6 +30,7 @@ const (
 	ConfigInfrastructureProviderPathName = "INFRASTRUCTURE_PROVIDER_CONFIG_FILEPATH"
 
 	ConfigShootName                 = "SHOOT_NAME"
+	ConfigShootAnnotations			= "SHOOT_ANNOTATIONS"
```
schrodit, line 33: go fmt

pkg/testmachinery/metadata/types.go
```diff
@@ -30,6 +30,7 @@ type Metadata struct {
 	OperatingSystemVersion    string `json:"operating_system_version,omitempty"`
 	Zone                      string `json:"zone,omitempty"`
 	AllowPrivilegedContainers *bool  `json:"allow_privileged_containers,omitempty"`
+	ShootAnnotations          string `json:"shoot_annotations,omitempty"`
```
schrodit, line 33: I think for the elasticserach serach it would be better if we had the real map available not just the string.
Then searching would be better.

pkg/common/common.go
dguendisch: right

pkg/testrun_renderer/templates/shoots.go
dguendisch: 👍

pkg/testmachinery/metadata/types.go
dguendisch: You're right, I was too lazy to add the respective unmarshal function to convert back from string to the map :)
Done now.

dguendisch:
Addressed and renamed the shoot flavor to `addtionalAnnotation` in the fields and just `annotation` in the flavor yaml.
ptal

PR #300, opened 2020-07-12
Add additional dimension info to slack table renderer to include NoPrivCtrs

dguendisch:
Adds a generic "additional dimension info" field to the slack output table to include additional dimension/metadata information like the recently introduced `allowPrivilegeContainers`.

### DIFF
--- a/pkg/testrunner/result/summary-poster.go
+++ b/pkg/testrunner/result/summary-poster.go
@@ -83,8 +83,12 @@ func parseTestrunsToTableItems(runs testrunner.RunList) (tableItems util.TableIt
 				}
 			}
 
+			var additionalDimensionInfo string
+			if !*meta.AllowPrivilegedContainers {
+				additionalDimensionInfo = "NoPrivCtrs"
+			}
 			item := &util.TableItem{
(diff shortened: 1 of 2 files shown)


dguendisch:
Hmm, that place gets the data from the github assets, hence the additional dimension information from the testruns is not present.
So if we'd want to have the info there, we'd need to amend the github assets with the new additional dimension information.
Is it worth the effort, wdyt?

schrodit:
> So if we'd want to have the info there, we'd need to amend the github assets with the new additional dimension information.
> Is it worth the effort, wdyt?

depend, if we need the differentiation in the uploaded github result and with that also in the corresponding slack notification. 
If we do not differentiate, it would just aggregate based on the other dimensions and will not distinguish between `allowPrivilegeContainers=on` and `allowPrivilegeContainers=off`.
But this is only a problem if one fails and we do not display the failed ones, which is not really likely to occur on the uploaded results.

dguendisch:
Then I suggest to proceed with this PR as is. If we need it at some point in time more formalized in the github assets, then we add it via a separate PR.

PR #305, opened 2020-07-24
revert testrunner watches to intelligent polling

schrodit:

Fixes #

**Release note**:

```improvement operator
A new polling informer for the testrun watch has been added.
```
```improvement operator
Testrunner watches do now use the new polling informer.
```

### DIFF
--- a/pkg/testmachinery/controller/watch/watch_test.go
+++ b/pkg/testmachinery/controller/watch/watch_test.go
@@ -16,6 +16,7 @@ package watch_test
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"testing"
 	"time"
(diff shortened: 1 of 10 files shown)


pkg/testmachinery/controller/watch/watch_test.go
```diff
@@ -109,117 +120,125 @@ var _ = Describe("Watch", func() {
...
-		w, err := watch.New(log.NullLogger{}, restConfig, nil)
-		Expect(err).ToNot(HaveOccurred())
-		go func() {
-			err := w.Start(ctx.Done())
-			Expect(err).ToNot(HaveOccurred())
-		}()
+				startTime := time.Now()
+				err = w.WatchUntil(2*time.Second, "test", "test", func(tr *tmv1beta1.Testrun) (bool, error) { return false, nil })
+				Expect(err).To(HaveOccurred())
+				Expect(time.Now().Sub(startTime).Seconds()).To(BeNumerically("~", 2, 0.01))
```
dguendisch, line 143: Let's see if that precision is not too strict and the test will flake (i.e. because of resource/cpu limits), so let's try :)

PR #306, opened 2020-07-31
Add "additionalLocations" to flavors

dguendisch:
Adds field `additionalLocations` to flavors. It allows to add or overwrite locations from where the actual tests are fetched.
Usage example:
```
flavors:
- provider: aws
  additionalLocations:
  - type: git
    repo:  https://github.com/gardener/gardener
    revision: master
```
**Release note**:

```improvement user
Flavors can customize created TestRuns with additional or overwritten testlocations with field `additionalLocations` containing an array of testlocation objects.
```

### DIFF
--- a/pkg/apis/testmachinery/v1beta1/types.go
+++ b/pkg/apis/testmachinery/v1beta1/types.go
@@ -15,6 +15,7 @@
 package v1beta1
 
 import (
+	"fmt"
 	argov1 "github.com/argoproj/argo/pkg/apis/workflow/v1alpha1"
 	"github.com/gardener/test-infra/pkg/util/strconf"
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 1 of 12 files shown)


pkg/testrunner/template/values.go
```diff
@@ -156,6 +156,9 @@ func (r *shootValueRenderer) GetValues(shoot *common.ExtendedShoot, defaultValue
 	if shoot.AdditionalAnnotations != nil {
 		values["shoot"].(map[string]interface{})["shootAnnotations"] = util.MarshalMap(shoot.AdditionalAnnotations)
 	}
+	if shoot.AdditionalLocations != nil && len(shoot.AdditionalLocations) > 0 {
```
schrodit, line 159: should we rather add the additional location like we do with the component descriptor?
So directly add them directly on the internal testrun instead of relying on helm. 
see https://github.com/gardener/test-infra/blob/master/pkg/testrunner/template/renderer.go#L93

With the current implementation we would also require every testrun to implement the `range` op in its charts.

WDYT?

dguendisch:
Tried another variant by directly injecting additional locations (instead of relying on templating) as you suggested, PTAL.

PR #309, opened 2020-09-17
Fix gardensetup test

schrodit:
- Only add shoot annotations if they are defined
- allow empty config elements

**Release note**:

```improvement operator
It is now allowed to specify empty config elements.
```


schrodit:
garden-setup is in maintenance all further test will be done using the landscaper setup.
Therefore its likely that this test will be removed in the near future. (so no fix/effort is needed in this direction)

PR #310, opened 2020-10-10
New elasticsearch command to update/rewrite data

dguendisch:
Adds a new command `elasticsearch` to update es data.
For now has only one subcommand `precompute`, which allows to rerun the computation of the precompute fields (stored in the teststep es documents in `.pre`) and update the data in elasticsearch.
Useful for modifying or introducing new pre fields and updating all existing data.

Fixes #

**Release note**:

```improvement user
A new command `elasticsearch precompute` was introduced to allow to update all teststep `.pre` fields in elasticsearch to the current/most up-to-date precompute values.
```

### DIFF
--- a/cmd/elasticsearch/cmd/precompute/precompute.go
+++ b/cmd/elasticsearch/cmd/precompute/precompute.go
@@ -1,0 +1,171 @@
+// Copyright 2020 Copyright (c) 2020 SAP SE or an SAP affiliate company. All rights reserved. This file is licensed under the Apache Software License, v. 2 except as noted otherwise in the LICENSE file.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package precompute
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"github.com/gardener/test-infra/pkg/apis/config"
+	"github.com/gardener/test-infra/pkg/logger"
+	"github.com/gardener/test-infra/pkg/testmachinery/collector"
+	"github.com/gardener/test-infra/pkg/util/elasticsearch"
+	"github.com/spf13/cobra"
+	"io"
+	"net/http"
+	"reflect"
+)
+
+var (
+	// only touch ES when true, dry-run otherwise
+	updateES bool
+)
+
+// AddCommand adds the precompute subcommand to another command.
+func AddCommand(cmd *cobra.Command) {
+	cmd.AddCommand(precomputeCmd)
+}
+
+var precomputeCmd = &cobra.Command{
+	Use:   "precompute",
+	Short: "Reads existing teststep metadata, re-computes the current precomputed values and optionally updates the respective elasticsearch document.",
+	PreRun: func(cmd *cobra.Command, args []string) {
+		if updateES {
+			logger.Log.Info("Starting 'elasticsearch precompute' in update mode", "elasticsearch endpoint", cmd.Flag("endpoint").Value, "elasticsearch user", cmd.Flag("user").Value)
+		} else {
+			logger.Log.Info("Starting 'elasticsearch precompute' in dry-run mode")
+		}
+	},
+	RunE: func(cmd *cobra.Command, args []string) error {
+		if err := run(cmd); err != nil {
+			logger.Log.Error(err, "error during execution")
+			return err
+		}
+		return nil
+	},
+	PostRun: func(cmd *cobra.Command, args []string) {
+		logger.Log.Info("Finished 'elasticsearch precompute'")
+	},
+}
+
+// package init defines the flags for the precompute command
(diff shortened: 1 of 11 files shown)


cmd/elasticsearch/cmd/precompute/precompute.go
```diff
@@ -0,0 +1,171 @@
...
+	if len(modifiedItems) == 0 {
+		logger.Log.V(6).Info("no modified data for any result (in this page) -> nothing to update")
+		return esResponse, nil
+	}
+
+	path, payload, err = BuildBulkUpdateQuery(modifiedItems)
+	if err != nil {
+		return QueryResponse{}, err
+	}
+	logger.Log.V(6).Info("going to bulk update", "path", path)
```
schrodit, line 155: shouldn't this be moved to line 163 to indicate the real update call?

cmd/elasticsearch/cmd/precompute/precompute.go
```diff
@@ -0,0 +1,171 @@
...
+
+
+	if !updateES {
+		logger.Log.Info("Not modifying elasticsearch data as update flag was not set")
```
schrodit, line 159: maybe logging the build path and payload would also help debugging when doing a dry run

pkg/util/elasticsearch/elasticsearch.go
```diff
@@ -68,11 +70,15 @@ func NewClient(cfg config.ElasticSearch) (Client, error) {
 }
 
 func (c *client) Request(httpMethod, rawPath string, payload io.Reader) ([]byte, error) {
-	esURL, err := c.parseUrl(rawPath)
+	return c.RequestWithCtx(context.Background(), httpMethod, rawPath, payload)
```
schrodit, line 73: this context should be canceled when the request is done
```suggestion
         ctx := context.Background()
         defer ctx.Done()
	return c.RequestWithCtx(ctx, httpMethod, rawPath, payload)
```

cmd/elasticsearch/cmd/precompute/precompute.go
```diff
@@ -0,0 +1,171 @@
...
+			}
+			if k8sVersion == "" && result.StepSummary.Metadata.Annotations != nil {
+				k8sVersion = result.StepSummary.Metadata.Annotations["testrunner.testmachinery.sapcloud.io/k8sVersion"]
+			}
+		}
+		oldPreComputed := result.StepSummary.PreComputed
+		if oldPreComputed != nil {
+			clusterDomain = oldPreComputed.ClusterDomain
+		}
+		newPreComputed := collector.PreComputeTeststepFields(result.StepSummary.Phase, k8sVersion, clusterDomain)
```
schrodit, line 135: you dont want to add a cluster domain if the old precomputed is not given.
Doesnt that mean that the cluster is only added if there are already precomputed fields?

pkg/testmachinery/collector/precompute.go
```diff
@@ -0,0 +1,68 @@
...
+			fmt.Printf("cannot parse k8s Version, will not precompute derived data: %s", err)
+		} else {
+			preComputed.K8SMajorMinorVersion = fmt.Sprintf("%d.%d", semVer.Major(), semVer.Minor())
+		}
+	}
+
+	if clusterDomain != "" {
+		preComputed.ArgoDisplayName = "argo"
+		preComputed.LogsDisplayName = "logs"
+		preComputed.ClusterDomain = clusterDomain
```
schrodit, line 64: the grafana and argo url are then computed like `argo.<clusterDomain>` ?

schrodit (changes requested):
just some minor observations

cmd/elasticsearch/cmd/precompute/precompute.go
```diff
@@ -0,0 +1,171 @@
...
+	})
+	if err != nil {
+		return err
+	}
+
+	pageSize := 10
+	path, payload := BuildScrollQueryInitial("testmachinery-*", pageSize)
+
+	var processedItems int
+	for i := 0;; i++ {
```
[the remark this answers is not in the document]
dguendisch: 👍

cmd/elasticsearch/cmd/precompute/precompute.go
dguendisch: You're right!

cmd/elasticsearch/cmd/precompute/precompute.go
dguendisch: Good idea, done.

pkg/testmachinery/collector/precompute.go
dguendisch: Not really, the displayname is just a "display name" and makes an artificial column in the dataset, which can be used to attach computed values (like concatenations) in kibana/grafana at runtime. But one needs that static artificial column.

cmd/elasticsearch/cmd/precompute/precompute.go
dguendisch: I struggled a bit to recompute that value in case this field is not already present (I'm not sure which data I can rely on for old datasets, so I don't want to set "some" clusterdomain there in case the information is not present).

pkg/util/elasticsearch/elasticsearch.go
dguendisch: Oh right, thanks for spotting, fixed.

dguendisch:
Addressed your comments, thanks for them! Also added some better error handling/reporting in case bucket updates contain failures.
I now also added the first new precomputed field: `providerEnhanced`.
PTAL :)

schrodit (approved):
/lgtm

PR #312, opened 2020-11-16
Feature/git env

schrodit:

Adds additional env vars for git location.
`GIT_SHA`: the commit sha of the specified location. This env var is always set.
`GIT_REF`: the specified `testLocation.Revision`. Is only set if the revision is not a commit sha

cc @timebertt 

@dguendisch also fixed the formatting of some files

**Release note**:

```improvement operator
Additional environment variables for `TM_GIT_SHA` and `TM_GIT_REF` have been added.
```

### DIFF
--- a/docs/testmachinery/GetStarted.md
+++ b/docs/testmachinery/GetStarted.md
@@ -88,6 +88,8 @@ spec:
 | TM_KUBECONFIG_PATH      | points to a directory containing all kubeconfig files (defaults to `/tmp/env/kubeconfig`). </br> The files contained in this dir depend on the concrete TestRun and can contain up to 5 files: <ul><li>_testmachinery.config_: the kubeconfig pointing to the testmachinery cluster</li><li>_host.config_: the kubeconfig pointing to the cluster hosting the gardener (not available for untrusted steps)</li><li>_gardener.config_: the kubeconfig pointing to the gardener cluster created by TM (or predefined/given by a TestRun, not available for untrusted steps)</li><li>_seed.config_: the kubeconfig pointing to the seed cluster configured by TM (or predefined/given by a TestRun, not available for untrusted steps)</li><li>_shoot.config_: the kubeconfig pointing to the shoot cluster created by TM (or predefined/given by a TestRun)</li></ul>|
 | TM_EXPORT_PATH | points to a directory where the test can place arbitrary test-run data which will be archived at the end. Useful if some postprocessing needs to be done on that data. Further information can be found [here](#export-contract) |
 | TM_TESTRUN_ID | Name of the testrun |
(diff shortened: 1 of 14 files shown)


dguendisch (approved):
Great, thanks!

pkg/testmachinery/config.go
```diff
@@ -71,6 +71,12 @@ const (
 	// TM_TESTRUN_ID_NAME is the name of the environment variable that holds the current testrun id
 	TM_TESTRUN_ID_NAME = "TM_TESTRUN_ID"
 
+	// TM_GIT_SHA_NAME is the name of the environment variable that holds the current git commit sha of the specified location.
+	TM_GIT_SHA_NAME = "GIT_SHA"
+
+	// TM_GIT_REF_NAME is the name of the environment variable that holds the current git ref of the specified location.
+	TM_GIT_REF_NAME = "GIT_REF"
```
timebertt, line 78: Any specific reason, why you didn't prefix the new env vars with `TM_`?

docs/testmachinery/GetStarted.md
```diff
@@ -88,6 +88,8 @@ spec:
...
 | TM_EXPORT_PATH | points to a directory where the test can place arbitrary test-run data which will be archived at the end. Useful if some postprocessing needs to be done on that data. Further information can be found [here](#export-contract) |
 | TM_TESTRUN_ID | Name of the testrun |
+| GIT_SHA | The commit SHA of the used git location |
+| GIT_REF | The ref of the used git location. Will be empty if only a commit sha is provided |
```
timebertt, line 92: Just to make it a bit more clear. If I would have read this without discussing it with you beforehand, I probably would have been confused.
```suggestion
| GIT_REF | The ref of the used git location (branch or tag name). Will be empty if only a commit sha is provided (e.g. when testing a Pull Request via the TM bot) |
```

timebertt (commented):
Thanks for adding this!
Just two minor comments / questions from my side.

pkg/testmachinery/config.go
schrodit: Not really. My idea was that I only want to sprefix testmachinery specific env vars with `TM` but git has not much todo with the testmachinery:
But I have no problem about changing it to `TM_GIT_xx`.
@dguendisch wdyt?

dguendisch: I think adding `TM_` is more consistent. I always viewed `TM_` vars as the TM's "API", as we define the presence and semantics.
Hence I'd vote for adding it.

schrodit:
@timebertt PTAL

timebertt (approved):
Very nice.
/lgtm

PR #317, opened 2020-12-01
Add the execution group id from the TestRun also to the TestSteps

dguendisch:
Adds the execution group id from the TestRun also to the TestSteps metadata.
This allows to filter in dashboards later on on the execution group id to see all teststeps from a respective execution group (i.e. one integration tests "run").

### DIFF
--- a/pkg/testmachinery/collector/summary.go
+++ b/pkg/testmachinery/collector/summary.go
@@ -19,6 +19,7 @@ import (
 	argov1 "github.com/argoproj/argo/pkg/apis/workflow/v1alpha1"
 	"github.com/gardener/gardener/pkg/utils"
 	tmv1beta1 "github.com/gardener/test-infra/pkg/apis/testmachinery/v1beta1"
+	"github.com/gardener/test-infra/pkg/common"
 	"github.com/gardener/test-infra/pkg/testmachinery/metadata"
 	"github.com/gardener/test-infra/pkg/util"
 	"github.com/gardener/test-infra/pkg/util/elasticsearch/bulk"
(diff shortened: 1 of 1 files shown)


pkg/testmachinery/collector/summary.go
```diff
@@ -97,6 +98,7 @@ func (c *collector) generateSummary(tr *tmv1beta1.Testrun, meta *metadata.Metada
 				stepMetadata.Configuration[elem.Name] = elem.Value
 			}
 		}
+		stepMetadata.Annotations[common.LabelTestrunExecutionGroup] = meta.Testrun.ExecutionGroup
```
schrodit, line 101: I think the execution group can be empty.
Do we then want to have a execution group annotation that is empty?

dguendisch: I think it can't be empty:
https://github.com/gardener/test-infra/blob/1ca8531272ffe13ed280a3c52009650c042c6809/pkg/testrunner/run.go#L31-L33 and 
https://github.com/gardener/test-infra/blob/ea986ed8f24ab178ebc9dfacfb4e3b32f5ba5785/pkg/testmachinery/metadata/metadata.go#L87
are the only setters.
But sure, it doesn't harm to check here as well, should I?

schrodit: With the new testrunner template flags I also added the possibility to not create a execution group.
Although, we do not use that flag together with the summary collector I would check it.
As you said, it does not harm

dguendisch: Makes sense in any case, you're right.

schrodit (approved):
/lgtm

PR #326, opened 2021-01-14
fix broken domain parsing regex

dguendisch:
Domain parsing got broken with https://github.com/gardener/test-infra/pull/325 -> fixed plus added a test for this.


schrodit (approved):
/lgtm

PR #328, opened 2021-01-17
switch to component descriptor v2

schrodit:

Switched the component descriptor to v2.
As the new component descriptor is distributed an oci client is needed to resolve transitive component descriptors

cc @ccwienk 

**Release note**:

```improvement operator
The testrunner now uses the component descriptor v2
```


schrodit:
unit tests are adjusted and tests but still need e2e testing
/hold

PR #329, opened 2021-01-19
Skip MetricsGrabber tests with >=1.20 (can't work)

dguendisch:
Skip MetricsGrabber tests. Since k8s v1.20.x these tests fail if they cannot find the control plane (kapi, kcm, ...) in `kube-system` ns. Of course these are not available in a shoot cluster, but for whatever reason the tests fail now if these pods are not found (probably a bug): https://github.com/kubernetes/kubernetes/commit/20863a8b4390d275925338571e7a46105177a6cf
So we'll skip them for us (they will anyway never work/test something useful).


schrodit (approved):
/lgtm

PR #333, opened 2021-04-07
improve testmachinery validation

schrodit:

The validation webhook performance has been improved by moving the time-consuming testdefinition validation (that includes lot of calls to github) to the default reconcile cycle.
The validation now only includes the validation of the testrun structure.

Note: currently the validation is not retried. The testrun is directly set to failed if the validation failed. 

Fixes #

I also updated the validation to use the default kubernetes validation helpers instead of our own identifiers.

**Release note**:

```improvement operator
The testmachinery validation has been improved by moving the time-consuming testdefinition validation to the default reconcile cycle.
The validation webhook now only validates the raw testrun without external resources.
```


pkg/shoot-telemetry/cmd/root.go
```diff
@@ -59,7 +59,7 @@ func GetRootCommand() *cobra.Command {
 					config.ShootsFilter[shoot] = true
 				}
 
-				// Validate the passed flag inputs.
+				// HasParent the passed flag inputs.
```
dguendisch, line 62: ???

dguendisch (changes requested):
Thanks! Only one nit

pkg/shoot-telemetry/cmd/root.go
schrodit: Oh yes must have been automatically refactored.

dguendisch (approved):

PR #336, opened 2021-04-19
add own ttl controller that deletes testruns after the ttl exceeds

schrodit:

Adds our own ttl controller that deletes testruns after their ttl has exceeded.
This was done to not rely on the argo ttl controller anymore.
The argo ttl controller and the testmachinery ttl controller do not interfere with each other so it is safe to run both controllers in parallel.

Fixes #

**Release note**:

```improvement operator
A TTL controller has been added that deletes testruns after their configured ttl time has exceeded.
```

### DIFF
--- a/pkg/apis/config/config.go
+++ b/pkg/apis/config/config.go
@@ -37,20 +37,31 @@ type Configuration struct {
 
 // Controller holds information about the testmachinery controller
 type Controller struct {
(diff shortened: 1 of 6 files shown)


dguendisch (approved):
thanks for the fast addition!!
/lgtm

PR #343, opened 2021-05-05
move three success test cases from skip.json to working.json and remove Not Found test case from skip.json

neo-liang-sap:
These three test cases have been validated success against aws/gcp/azure on k8s 1.20.4, hence moving it from skip.json to working.json
Also there are some test case Not Found , remove them from skip.json
Fixes #

@dguendisch @schrodit FYI currently SRE team breaks down all test cases into several sub-tasks, e.g. in skip.json there are totally 676 cases, i'm working on the 600-676 line, so this PR only for adding/removing test cases from 600-676, other people in SRE team will submit other PRs correspondingly
**Release note**:

```improvement operator

```


integration-tests/e2e/kubetest/description/1.20/working.json
```diff
@@ -165,6 +166,8 @@
   { "testcase": "[sig-storage] In-tree Volumes [Driver: ceph][Feature:Volumes] [Testpattern: Pre-provisioned PV (default fs)] subPath should support readOnly file specified in the volumeMount", "groups": ["fast"], "only": ["openstack"] },
   { "testcase": "[sig-storage] In-tree Volumes [Driver: ceph][Feature:Volumes] [Testpattern: Pre-provisioned PV (default fs)] volumes should allow exec of files on the volume", "groups": ["fast"], "only": ["openstack"] },
   { "testcase": "[sig-storage] In-tree Volumes [Driver: ceph][Feature:Volumes] [Testpattern: Pre-provisioned PV (default fs)] volumes should be mountable", "groups": ["fast"], "only": ["openstack"] },
+  { "testcase": "[sig-storage] CSI Volumes [Driver: csi-hostpath] [Testpattern: Dynamic PV (default fs)] subPath should verify container cannot write to subpath readonly volumes [Slow]","groups": ["slow"] },
+  { "testcase": "[sig-storage] PersistentVolumes NFS when invoking the Recycle reclaim policy should test that a PV becomes Available and is clean after the PVC is deleted.", "groups": ["fast"] },
```
schrodit, line 170: do you know how long these tests run?
I think we said that the `fast` group only contains tests that are faster than 2min (if I remember correctly)

neo-liang-sap: hmm, i need to confirm this by running them in single test case - so if the time is longer than 2 mins but not very long, it will be put in `slow` group?

schrodit: yes. that is because we wanted the fast group to be as fast as possible. 
Because even 2-min tests will accumulate when we run hundreds.

neo-liang-sap: hmm, actually these three test cases are less than 2 mins  (the longest one is 88 seconds)
image
however i saw one of them has `slow` in title so i put it in slow - so shall i change these three all to `fast` group?

schrodit: if all are max 88sec I would put them into the fast group.
@dguendisch any objections?

dguendisch: fine with me

dguendisch (approved):
/lgtm

schrodit (approved):

PR #345, opened 2021-05-10
remove not found #411

tedteng:
remove not found tests from #411

Fixes 

**Release note**:

```improvement operator

```


integration-tests/e2e/kubetest/description/1.20/skip.json
```diff
@@ -1,101 +1,12 @@
-[{ "testcase": "[Flaky]"},
-  { "testcase": "Alpha"},
-  { "testcase": "[Experimental]"},
```
schrodit, line 3: I think we should keep this list of tests that are skipped in general.

integration-tests/e2e/kubetest/description/1.20/skip.json
```diff
@@ -1,101 +1,12 @@
...
-  { "testcase": "[Driver: vSphere]"},
-  { "testcase": "[Driver: cinder]"},
-  { "testcase": "[Driver: gluster]"},
```
schrodit, line 6: see comment above

schrodit (changes requested):
otherwise lgtm

integration-tests/e2e/kubetest/description/1.20/skip.json
dguendisch: Our openstack provisioner is cinder, hence we could remove the cinder skip to have these test executed.
We can safely remove gluster however as we don't run/deploy it.

tedteng: if I understand correctly remove `cinder` and keep `gluster` in skip is that correct ?

dguendisch: yes (removing cinder from skip will result in executing cinder tests -> we want that as we do have cinder csi in place; keeping gluster in skip will result in skipping gluster tests -> we want that as we have no interest in gluster)

dguendisch (approved):
/lgtm

PR #363, opened 2021-08-19
Update alpine 3.13 to 3.14

dguendisch:

/area open-source
/kind cleanup

Updates base images from alpine 3.13 to 3.14 (which also updates python 3.8 to 3.9) so that newer versions of cicd libs can be installed.

/invite @schrodit
cc @vpnachev 

Fixes #

**Release note**:

```other operator

```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -21,7 +21,7 @@ COPY . .
 RUN make install
 
 ############# tm-controller #############
-FROM alpine:3.13 AS tm-controller
+FROM alpine:3.14 AS tm-controller
 
 COPY charts /charts
 COPY --from=builder /go/bin/testmachinery-controller /testmachinery-controller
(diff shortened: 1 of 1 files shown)


schrodit (approved):

vpnachev:
If you can ensure this image will run on new enough container runtime (see https://wiki.alpinelinux.org/wiki/Release_Notes_for_Alpine_3.14.0#faccessat2), I am fine, but I would suggest to keep running on 3.13 for some more months.

dguendisch:
Turns out we do have runtimes with not sufficiently new runc version, hence alpine 3.14 is not an option yet...
/hold

dguendisch:
> Turns out we do have runtimes with not sufficiently new runc version, hence alpine 3.14 is not an option yet...

These runtimes are now upgraded to recent enough runc/docker/libseccomp versions to be able to use alpine 3.14.
/unhold

dguendisch:
cc @ccwienk @jkrayl

schrodit (approved):
/lgtm

PR #364, opened 2021-08-19
Switch base from alpine to debian

dguendisch:

/area open-source
/kind cleanup

Switches the base image from alpine to debian bullseye, allows to combine go 1.16 with python 3.9 => image size increased from (already huge?) 1.7GB to 2.0GB, image build time decreased from ~350secs to ~50secs because pip packages no longer need to build their deps from source (because of musl vs. glibc).
Not sure if it would be a good idea though, hence just a draft.

cc @schrodit @vpnachev 

Fixes #

**Release note**:

```other operator

```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -43,42 +43,29 @@ WORKDIR /
 ENTRYPOINT ["/telemetry-controller"]
 
 ############# tm-base-step #############
-FROM golang:1.16-alpine3.13 AS base-step
+FROM golang:1.16-bullseye AS base-step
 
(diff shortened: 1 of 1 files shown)


vpnachev:
I was about to propose to use `apk add --no-cache` and `rm -rf /var/lib/apt/lists/*` after `apt-get install` to reduce the image size, but running it locally the image is still 2.13 GiB

vpnachev:
Not sure what is the state of this PR, having mind it is ~2 years old now, maybe we can close it?

PR #366, opened 2021-08-20
Improve ns handling and inc test timeout tolerance

### LINKED ISSUE #365: [Flaky Test] controller_test
**How to categorize this issue?**

/area testing
/kind flake
/priority 3

**Which test(s)/suite(s) are flaking**:
https://github.com/gardener/test-infra/blob/f494cd53f4571bc4e4d3d0bbc9f81dc379420668/pkg/testmachinery/controller/ttl/controller_test.go

**CI link**:
https://concourse.ci.gardener.cloud/teams/gardener/pipelines/test-infra-master/jobs/master-pull-request-job/builds/67

**Reason for failure**:
A test namespace is actually composed of a rand int from 0 to 1000, hence with some tests it collides here and there.
However upon checking I noticed that the `AfterEach` https://github.com/gardener/test-infra/blob/f494cd53f4571bc4e4d3d0bbc9f81dc379420668/pkg/testmachinery/controller/ttl/controller_test.go#L64-L72 doesn't really make sense: it seems this should wait until a namespace is deleted but it happily progresses when the namespace is simply existing and in `Terminating` state as this won't throw any error...
I think that the namespace will actually never delete as there is probably no gc controller running in the testenv that would also terminate the contained resources like `TestRun`.

@schrodit do I miss another intent? If not, either testenv can be setup with a gc controller or we remove the resp. wait for ns deletion and instead increase the randomness greatly.

**Anything else we need to know**:

dguendisch:

/area testing
/kind enhancement

Improve ns generation to avoid collision (happened to me once) and increases the timeout tolerance of one test (happened to me sometimes).
Removes the unnecessary (and wrong) ns deletion from `afterEach`:
- according to https://cluster-api.sigs.k8s.io/developer/testing.html#envtest:
```
One common example is garbage collection; because there are no controllers monitoring built-in resources, objects do not get deleted, even if an OwnerReference is set up; as a consequence, usually test implements code for cleaning up created objects.
```

/invite @schrodit 

Fixes #365

**Release note**:

```other operator

```

### DIFF
--- a/pkg/testmachinery/controller/ttl/controller_test.go
+++ b/pkg/testmachinery/controller/ttl/controller_test.go
@@ -16,8 +16,6 @@ package ttl_test
 
 import (
 	"context"
-	"fmt"
-	"math/rand"
 	"time"
 
 	"github.com/go-logr/logr"
@@ -45,7 +43,7 @@ var _ = Describe("Reconcile", func() {
 	BeforeEach(func() {
 		ctx = context.Background()
 		ns := &corev1.Namespace{}
-		ns.Name = fmt.Sprintf("test-%d", rand.Intn(1000))
+		ns.GenerateName = "test-"
 		if err := fakeClient.Create(ctx, ns); err != nil {
 			if !apierrors.IsAlreadyExists(err) {
 				Expect(err).ToNot(HaveOccurred())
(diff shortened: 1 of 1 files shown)


schrodit (approved):
/lgtm

PR #371, opened 2021-09-09
test case 3501 - 4128

jguipi:
/area testing
/kind TODO

- test case 3501 - 4128

Fixes #

**Release note**:

- test case 3501 - 4128


jguipi:
The following test case seems to be falling every time:

```
[
    { "testcase": "[sig-storage] CSI Volumes [Driver: csi-hostpath] [Testpattern: Dynamic PV (default fs)] subPath should unmount if pod is gracefully deleted while kubelet is down [Disruptive][Slow][LinuxOnly]"},
    { "testcase": "[sig-storage] In-tree Volumes [Driver: local][LocalVolumeType: blockfs] [Testpattern: Pre-provisioned PV (default fs)] subPath should unmount if pod is gracefully deleted while kubelet is down [Disruptive][Slow][LinuxOnly]"},
    { "testcase": "[sig-scheduling] SchedulerPriorities [Serial] PodTopologySpread Scoring validates pod should be preferably scheduled to node which makes the matching pods more evenly distributed"},
    { "testcase": "[sig-storage] In-tree Volumes [Driver: local][LocalVolumeType: blockfs] [Testpattern: Pre-provisioned PV (default fs)] subPath should unmount if pod is gracefully deleted while kubelet is down [Disruptive][Slow][LinuxOnly]"},
    { "testcase": "[sig-storage] In-tree Volumes [Driver: local][LocalVolumeType: dir-bindmounted] [Testpattern: Pre-provisioned PV (default fs)] subPath should unmount if pod is gracefully deleted while kubelet is down [Disruptive][Slow][LinuxOnly]"},
    { "testcase": "[sig-storage] In-tree Volumes [Driver: local][LocalVolumeType: dir-link-bindmounted] [Testpattern: Pre-provisioned PV (default fs)] subPath should unmount if pod is gracefully deleted while kubelet is down [Disruptive][Slow][LinuxOnly]"},
    { "testcase": "[sig-storage] In-tree Volumes [Driver: nfs] [Testpattern: Dynamic PV (filesystem volmode)] disruptive[Disruptive][LinuxOnly] Should test that pv written before kubelet restart is readable after restart."}
]
```

jguipi:
> I think we can merge, just a final question: did you test these tests with the cloudproviders we support?

yes !  
If I'm not wrong we are not supporting CSI driver. but all the other have been test in all cloudprovider

dguendisch:
> If I'm not wrong we are not supporting CSI driver. but all the other have been test in all cloudprovider

We did analyze several CSI related issues in various Hands-On sessions, so certainly we do have CSI enabled nearly everywhere ;)
Could you check why the above tests fail?

PR #374, opened 2021-10-11
remove resource manager

hendrikKahl:

/area testing
/kind cleanup
/kind technical-debt

Remove the Gardener Resource Manager, which so far has managed the deployment of dependent helm charts. This reduces the project's dependencies significantly. All charts previously stored in `charts/internal` have been integrated into the sub-chart structure of `testmachinery` .

The validation webhook has been extended to check the availability of the argo workflow controller deployment periodically. 

Fixes #

**Release note**:

```other operator
Removal of Gardener resource manager and refactoring of the helm chart structure. Testmachinery chart deploys all of its dependencies directly.
```


charts/testmachinery/charts/argo/templates/config.yaml
```diff
@@ -28,21 +28,24 @@ data:
...
+    {{- if .Values.argo.logging.additionallinks }}
+    {{- range .Values.argo.logging.additionallinks }}
     - name: {{ .name }}
       scope: {{ .scope }}
       url: {{ .url }}
-    {{ end }}
-    {{ end }}
+    {{- end }}
+    {{- end }}
+    {{- end }}
```
schrodit, line 41: is the the resulting yaml fine with argo if `argo.logging` is defined but not `argo.logging.links`?

charts/testmachinery/charts/argo/values.yaml
```diff
@@ -21,16 +21,21 @@ argo:
...
   # limits:
   #   cpu: 1
   #   memory: 1Gi
 
+
   logging:
-    links: []
+    # default links to a grafana instance will be injected
+    grafana:
+      host: "grafana.example.com"
```
schrodit, line 36: not sure if we should default the values to something that does not exist.
I would rather comment it and only set if a host really available.

charts/testmachinery/local-values.yaml
```diff
@@ -12,6 +12,14 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+global:
+  s3Configuration:
+    server:
+      ssl: false
+      endpoint: "tbd"
```
schrodit, line 19: her should also rather omit it like 
```suggestion
      # endpoint: "tbd"
```
and then use the helm `required` keyword to force that valid values are set.

We should do the same for the other s3Config Values.
With that we would get early feedback about missing configuration

charts/testmachinery/values.yaml
```diff
@@ -12,7 +12,14 @@
...
 
-
+global:
+  s3Configuration:
+    server:
+      endpoint: ""
+      ssl: true
+    bucketName: testmachinery
+    accessKey: IbnGHa4V6ypsm
+    secretKey: l9rJ0XR65Rkvc9g9fyOf
```
schrodit, line 22: see local values comment

pkg/apis/config/v1beta1/config.go
```diff
@@ -81,18 +80,12 @@ type ElasticSearch struct {
...
-	Enabled bool   `json:"enabled"`
-	Host    string `json:"host"`
+// HealthCheckTarget specifies a deployment whose health should be checked.
+type HealthCheckTarget struct {
+	//Namespace specifies the namespace where resources relevant for a health check exist in.
+	Namespace string `json:"namespace,omitempty"`
+	// DeploymentName is the name of a deployment whose health will be checked.
+	DeploymentName string `json:"deploymentName,omitempty"`
+	//IntervalSeconds specifies the frequency of the health check in seconds
+	IntervalSeconds int `json:"intervalSeconds,omitempty"`
```
schrodit, line 90: maybe we should use the k8s `metav1.Duration` field.
With that we could use human readable strings like `5m`

pkg/testmachinery/controller/admission/webhooks/validator.go
```diff
@@ -28,6 +36,44 @@ import (
...
 )
 
+var (
+	healthMutex sync.Mutex
+	healthErr   error
+)
+
+// StartHealthCheck will start a go routine periodically checking the health of a specified deployment
+// The result of the checks is available to the testRunValidator only
+func StartHealthCheck(ctx context.Context, reader client.Reader, namespace string, deplyomentName string, interval int) {
```
schrodit, line 46:
```suggestion
func StartHealthCheck(ctx context.Context, reader client.Reader, namespace string, deploymentName string, interval int) {
```

pkg/testmachinery/controller/admission/webhooks/validator.go
```diff
@@ -48,6 +94,13 @@ func (v *testrunValidator) InjectDecoder(d *admission.Decoder) error {
 
 // Handle validates a testrun
 func (v *testrunValidator) Handle(ctx context.Context, req admission.Request) admission.Response {
+	healthMutex.Lock()
+	if healthErr != nil {
+		defer healthMutex.Unlock()
+		return admission.Denied(healthErr.Error())
```
schrodit, line 100: I think we should rather error with `admission.Errored(http.InternalError, err)`.
Maybe something more suitable than `internalError`.
Because the testrun itself is not necessary wrong. but we some internal dependency errors

schrodit (changes requested):

charts/testmachinery/charts/argo/values.yaml
hendrikKahl: Tricky question - so far those "default" logging links have been derived automatically by the dependency ensurer: https://github.com/gardener/test-infra/blob/master/pkg/testmachinery/controller/dependencies/argo.go#L99 
However I would agree, that we should make this optional. I've changed the chart structure slightly to accommodate it.
Unfortunately I cannot implement a check similar to the ensurer since the `lookup` function returns nil when we use `helm template` to render the chart for `kubectl apply`.

pkg/testmachinery/controller/admission/webhooks/validator.go
hendrikKahl: HTTP 424 seems to be suitable.

go.mod
```diff
@@ -5,16 +5,21 @@ go 1.16
...
+	github.com/containerd/containerd v1.4.3 // indirect
+	github.com/containerd/continuity v0.0.0-20201208142359-180525291bb7 // indirect
+	github.com/docker/cli v20.10.3+incompatible // indirect
 	github.com/emicklei/go-restful v2.11.1+incompatible // indirect
 	github.com/fsnotify/fsnotify v1.4.9
 	github.com/gardener/component-cli v0.28.0
 	github.com/gardener/component-spec/bindings-go v0.0.52
 	github.com/gardener/gardener v1.19.1
-	github.com/gardener/gardener-resource-manager v0.22.0
 	github.com/gardener/gardener-resource-manager/api v0.22.0
```
schrodit, line 23: I thought that we might get rid of this dependency too?

pkg/apis/config/v1beta1/config.go
```diff
@@ -81,18 +80,12 @@ type ElasticSearch struct {
...
-	Enabled bool   `json:"enabled"`
-	Host    string `json:"host"`
+// HealthCheckTarget specifies a deployment whose health should be checked.
+type HealthCheckTarget struct {
+	//Namespace specifies the namespace where resources relevant for a health check exist in.
+	Namespace string `json:"namespace,omitempty"`
+	// DeploymentName is the name of a deployment whose health will be checked.
+	DeploymentName string `json:"deploymentName,omitempty"`
+	//IntervalSeconds specifies the frequency of the health check in seconds
+	Interval metav1.Duration `json:"interval,omitempty"`
```
schrodit, line 90: we should default this value or do not run any healthcheck if `Interval == 0`.
otherwise we would ddos the apiserver with health requests.

nit: the docstring has to be adjusted `/IntervalSeconds/Interval`

schrodit (approved):

PR #383, opened 2021-12-15
Update the go version of several TestDefinitions

ialidzhikov:
/area testing
/kind bug

The current HEAD of test-infra requires at least go1.15.
`make install` fails with the following error on go1.13 and go1.14:
```
$ docker run --rm -v ${GOPATH}/src:/go/src -it golang:1.14 bash
root@8f6f98ad9fa1:/go# cd src/github.com/gardener/test-infra/
root@8f6f98ad9fa1:/go/src/github.com/gardener/test-infra# make install
# golang.org/x/net/http2
vendor/golang.org/x/net/http2/transport.go:417:45: undefined: os.ErrDeadlineExceeded
^Cmake: *** [Makefile:72: install] Interrupt
```

Hence, TestDefinitions that are still on go1.13 fail with the same on test execution.

`os.ErrDeadlineExceeded` is introduced in go1.15 with https://cs.opensource.google/go/go/+/d422f54619b5b6e6301eaa3e9f22cfa7b65063c8.

**Release note**:

```other operator
The golang version of several TestDefinition is updated to match the required golang version by test-infra.
```


dguendisch (approved):
/lgtm
thank you very much!

PR #386, opened 2022-01-17
Update Gardener related dependencies

hendrikKahl:

/area testing
/kind technical-debt

Update Gardener related go dependencies to their latest version. Additionally, some components which still used `v1beta1` api groups have been switched to use `v1`. 
 
In order to ensure compatibility with k8s 1.19.x clients, some more components are pinned to a specific version. This can be changed/upgraded as soon as we move to 1.20 or higher.

Some linter issues where fixed that where reported by a newer version of the tool.

Fixes #

**Release note**:

```other operator
Dependency updates and fixed linter issues
```


Makefile
```diff
@@ -79,7 +79,7 @@ all: generate format verify install
 
 .PHONY: install-requirements
 install-requirements:
-	@curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b $(go env GOPATH)/bin v1.41.1
+	@curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b $(go env GOPATH)/bin v1.41.0
```
dguendisch, line 82: Why do we go back one patch version?

dguendisch (approved):
/lgtm

PR #387, opened 2022-01-26
add networkingType to extendedShootFlavor

### LINKED ISSUE #377: Support selecting the network type calico<->cilium
Testmachinery should support switching the [network type via the shoot spec](https://github.com/gardener/gardener/blob/master/example/90-shoot.yaml#L287) to allow tests to be created that don't target the default (calico) but can target cilium as well.
- [ ] Testmachinery flavors need to support this
- [x] [CreateShoot step](https://github.com/gardener/gardener/blob/master/.test-defs/CreateShoot.yaml) in g/g needs to support this

hendrikKahl:

/area testing
/kind enhancement

Add `NETWORKING_TYPE` to the list of things that can be specified per flavor. This will allow testing of a dedicated network component.

Fixes #377 

**Release note**:

```other user
Add an option to configure NETWORKING_TYPE per flavor.
```


dguendisch (approved):
/lgtm

PR #388, opened 2022-01-26
add a target to build test definition tools

hendrikKahl:

/area testing
/kind test

Add a `make` target + scripts to compile tools used in tests as defined in `.test-definitions`. A similar check is currently not available and thus basic installation issues are only detected when a test is executed.

As part of a recent dependency update, the go environment for these tests needs to be at least 1.16.

Fixes #

Once the PR is merged, we can add a step to the pipeline to run this regularly:  https://github.com/gardener/test-infra/blob/36f2a9c57f8c9f3c85ecbfa2583bf5babcbd2ce0/branch.cfg#L52-L56 

**Release note**:

```other operator
make target to install test-definitions to detect issues early.
```


dguendisch (approved):
/lgtm

PR #392, opened 2022-02-10
increase promtail memory limit

hendrikKahl:

/area testing
/kind impediment

Increase the memory limit of promtail. The current setting causes promtail pods to be oom-killed occasionally, when there are lots of pods on a larger node.

Fixes #

**Release note**:

```other operator
Increase promtail's memory limit.
```


dguendisch (approved):
/lgtm

PR #397, opened 2022-03-09
use workflow namespace for pod garbage collection

hendrikKahl:

/area testing
/kind bug

The testmachinery controller implements some basic garbage collection to clean-up pods related to completed workflows. Even though the controller deploys workflows into different namespaces, it tries to find any pod in its own namespace. 

This PR changes the behavior to look into the namespace of the actual workflow for garbage collectoin.

Fixes #

/invite @dguendisch 

**Release note**:

```other operator
Fix pod garbage collection when using multiple namespaces
```


dguendisch (approved):
/lgtm

PR #398, opened 2022-03-22
Add OpenVPN to base instance

briantopping:
Add OpenVPN to allow connection to Google Cloud VMware Engine instances (https://github.com/gardener/gardener-extension-provider-vsphere/issues/221)

Signed-off-by: Brian Topping <brian.topping@sap.com>

/area dev-productivity
/kind enhancement

GCVE instances require OpenVPN for VMware cluster access. Adding the client to the base image will allow tests to set up a tunnel. for access to ESX/NSX.

Fixes #

**Release note**:

```other operator

```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -70,6 +70,7 @@ RUN  \
     openssh \
     openssl \
     openssl-dev \
+    openvpn \
     python3 \
     python3-dev \
     py3-pip \


dguendisch:
Hi Brian, I prefer to not introduce additional dependencies from external tests to this image as we can accidentally break you when updating this image or have conflicting needs.
You are free to use any image that suits you to execute your tests. See how gardener is using i.e. just the golang image: https://github.com/gardener/gardener/blob/master/.test-defs/TestSuiteShootDefault.yaml#L25

briantopping:
Got it, thanks Dieter! I fully understand the distinction now, thanks for explaining it! 👍

PR #403, opened 2022-04-07
Don't add ValidatingWebhookConfiguration in chart local mode

briantopping:
Don't add `ValidatingWebhookConfiguration` object if `.Values.testmachinery.local` is set. `local` mode causes the admission controller endpoint to not be added.

Signed-off-by: Brian Topping <brian.topping@sap.com>

/area testing
/kind bug


hendrikKahl (approved):
good catch, thx!

PR #405, opened 2022-04-19
Render all test results if multiple results exist for a single cell

dguendisch:

/area testing
/kind bug

Render all test results if multiple results exist for a single key, i.e. there were was a succesfull result for GCP+1.21+GL576 but a failed one for GCP+1.21+GL387 and the result matrix has only one line for GCP+1.21.
Does not add new dimensions for now (but can be added later), but just ensures that all results are visible.
Previously only one of the two results was rendered.
With this fix all are rendered:
```
      +---+--------+
      |   |  GCP   |
      +---+--------+
      |   |  ✅ ❌ |
      +---+--------+
```

Fixes #

/invite @hendrikKahl 

**Release note**:

```other operator

```

### DIFF
--- a/pkg/testrunner/result/summary-poster_test.go
+++ b/pkg/testrunner/result/summary-poster_test.go
@@ -1,0 +1,58 @@
+package result
+
+import (
+	tmv1beta1 "github.com/gardener/test-infra/pkg/apis/testmachinery/v1beta1"
+	"github.com/gardener/test-infra/pkg/testmachinery/metadata"
+	"github.com/gardener/test-infra/pkg/testrunner"
+	"github.com/gardener/test-infra/pkg/util"
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+)
+
+var _ = Describe("summary-poster", func() {
+
+	It("should render summary with 1 and 2 results for the same line", func() {
+		run := &testrunner.Run{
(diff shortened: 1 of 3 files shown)


pkg/testrunner/result/summary-poster_test.go
```diff
@@ -0,0 +1,58 @@
...
+	"github.com/gardener/test-infra/pkg/testmachinery/metadata"
+	"github.com/gardener/test-infra/pkg/testrunner"
+	"github.com/gardener/test-infra/pkg/util"
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+)
+
+var _ = Describe("summary-poster", func() {
+
+	It("should render summary with 1 and 2 results for the same line", func() {
```
hendrikKahl, line 14: could this be split into 2 separate tests? If the fails, it would be more obvious where the problem is located.

hendrikKahl (changes requested):
Very nice, thanks a lot! 
I've added a few comments inline.

hendrikKahl:
as part of the concourse job, `make check` failed with an error saying ` summary-poster_test.go is unformatted`

dguendisch:
Applied your suggestions, PTAL.

hendrikKahl (approved):
/lgtm

PR #416, opened 2022-08-03
Raise default from 50 to 200 elements before pagination starts

dguendisch:

/area usability
/kind enhancement

Raises the default amount of entries in the tm-bot UI before pagination happens from `50` to `200`.
Background: We regularly have some > 100 testruns and the UI does not slow down with a few hundred entries. Hence no need for pagination in such cases.

Fixes #

**Release note**:

```other operator

```


dguendisch (commented):
/lgtm

hendrikKahl (approved):
/lgtm

PR #418, opened 2022-08-16
Adapt publish-tm-chart to work with helm v3.8 or later

hendrikKahl:
Signed-off-by: hendrikKahl <hendrik.kahl@sap.com>

/area testing
/kind technical-debt

Due to some changes in the concourse setup, the release job for this repo fails to publish the helm chart.
This PR adapts the scripts / make targets to work with helm v3 or later:
- stop downloading helm but assume a helm binary >= v3.8.0 will be present within it's concourse execution context
- rename the chart to `testmachinery-controller` as the metadata is used during upload. This way everything stays as it was from an external perspective and we avoid potentially breaking things
- use `package` and `push` to continue publishing the chart as an oci artifact 

Fixes #

/invite @dguendisch 

**Release note**:

```other operator
NONE
```

### DIFF
--- a/.ci/publish-helm
+++ b/.ci/publish-helm
@@ -31,13 +31,8 @@ else
   export SOURCE_PATH="$(readlink -f "${SOURCE_PATH}")"
 fi
 
-export DESIRED_VERSION=v3.2.1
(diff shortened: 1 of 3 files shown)


dguendisch (approved):
/lgtm

Issue #419, opened 2022-08-18
Support more version patterns in flavor configs


dguendisch:
Today we support things like
```
kubernetes:
    pattern: "latest"
```
(see https://github.com/gardener/test-infra/blob/master/docs/testrunner/README.md#shoot-flavor-configuration)

Sometimes we want to express a flavor that should select one minor version older than the latest one, something like `latest - 0.1`.
If arithmetic is too complex, we could also simply introduce a few fixed named patterns like `oneMinorBeforeLatest`, `twoMinorBeforeLatest`, ... (as we usually anyway only support a limited set of minor K8s versions at a time).

PR #420, opened 2022-08-19
Support version patterns to go back 1,2,3 or 4 minor versions

dguendisch:

/area testing
/kind enhancement

K8s version patterns in flavors now support going a few minors back __relative__ to the latest supported K8s versions.
This allows to define test flavors that can test older K8s versions without regularly updating the flavor files.
Can be used with or without `filterPatchVersions: true` depending on whether you only want the latest patch version of some old minor version or all patch versions of this old minor version.

Fixes https://github.com/gardener/test-infra/issues/419

/invite @hendrikKahl 

**Release note**:

```other user
Test flavor K8s version patterns in addition to `latest` now support `oneMinorBeforeLatest`, ... up to `fourMinorBeforeLatest`. Can be used with or without `filterPatchVersions: true`.
```

### DIFF
--- a/docs/testrunner/README.md
+++ b/docs/testrunner/README.md
@@ -84,7 +84,7 @@ For example: 3 kubernetes versions and 4 worker pools will result in 12 differen
 flavors:
 - provider: aws|gcp|azure|alicloud|openstack|...
   kubernetes:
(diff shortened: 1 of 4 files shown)


hendrikKahl (approved):
/lgtm

PR #428, opened 2022-11-22
Add version pattern unit test

dguendisch:

/area testing

Add version pattern unit test to clarify semantics of e.g. K8s version `>1.25`.

/invite @hendrikKahl 

Fixes #

**Release note**:

```other operator

```

### DIFF
--- a/pkg/util/kubernetes_version_test.go
+++ b/pkg/util/kubernetes_version_test.go
@@ -264,6 +264,32 @@ var _ = Describe("kubernetes version util", func() {
 			))
 		})
 
+		It("should return something (real) greater than 1.14", func() {
+			pattern := ">1.14"
+			versionFlavor := common.ShootKubernetesVersionFlavor{
+				Pattern: &pattern,
+			}
+
+			versions, err := util.GetK8sVersions(cloudprofile, versionFlavor, true)
+			Expect(err).ToNot(HaveOccurred())
+			Expect(versions).To(ConsistOf(
+				newExpirableVersion("1.15.2"),
+			))
+		})
+
+		It("should return versions in between", func() {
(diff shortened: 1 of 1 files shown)


shafeeqes (approved):
/lgtm

pkg/util/kubernetes_version_test.go
```diff
@@ -264,6 +264,41 @@ var _ = Describe("kubernetes version util", func() {
...
+			versionFlavor := common.ShootKubernetesVersionFlavor{
+				Pattern: &pattern,
+			}
+
+			versions, err := util.GetK8sVersions(cloudprofile, versionFlavor, true)
+			Expect(err).ToNot(HaveOccurred())
+			Expect(versions).To(ConsistOf(
+				newExpirableVersion("1.15.2"),
+			))
+			Expect(versions).To(Not(ConsistOf(
```
shafeeqes, line 278: Nit: We don't need to check for the Not case, since Consist passes only if it matches exactly the elements passed.
https://github.com/gardener/test-infra/blob/3f0c0249d5b005d666579a68cdd40b7a17531181/vendor/github.com/onsi/gomega/matchers.go#L298

If we still want to make the tests explicit, we can use `ContainElements` and `NotTo(ContainElements`
https://github.com/gardener/test-infra/blob/3f0c0249d5b005d666579a68cdd40b7a17531181/vendor/github.com/onsi/gomega/matchers.go#L319-L331

dguendisch: Thanks! I resorted to removing the unnecessary `...To(Not(...` parts.

hendrikKahl (approved):
/lgtm

PR #427, opened 2022-11-16
Adapt to kubetests format for K8s 1.25

dguendisch:
**How to categorize this PR?**

/area testing
/kind enhancement

**What this PR does / why we need it**:
Adapt the e2e test runner to be able to parse newer kubetest result formats for K8s 1.25 clusters and fix kubetest invocations for conformance tests which no longer worked for 1.25, instead use for 1.25 the suggested format of https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/conformance-tests.md#running-conformance-tests-with-kubetest.

/invite @hendrikKahl 
**Which issue(s) this PR fixes**:
Fixes errors like
```
couldn't unmarshal junit.xml /tmp/e2e/artifacts/1668608047/junit_01.xml: expected element type <testsuite> but have <testsuites>
```

**Special notes for your reviewer**:

**Release note**:

```other operator

```

PR #431, opened 2022-12-12
Fix invocation of conformance tests with `GINKGO_PARALLEL=false`

dguendisch:

/area testing
/kind bug

Fixes a bug with the invocation of conformance tests when `GINKGO_PARALLEL=false` is used. It was introduced with https://github.com/gardener/test-infra/pull/427 and results in conformance tests only executing the serial tests without everything else.

Fixes #

**Release note**:

```other operator

```


hendrikKahl (approved):
/lgtm

PR #433, opened 2022-12-14
Ensure "normal" testgrid jobs have no flakes

dguendisch:

/area testing
/kind enhancement

Ensures our serial execution of testgrid conformance jobs have no flakes. There's no point in executing them with flake tolerance while we don't tolerate flakes on cncf submission.

Changes a magic defaulting on the way where wrong startup flags were silently replaced with sane defaults to fail-fast instead.

/invite @hendrikKahl 

Fixes #

**Release note**:

```other operator

```


hendrikKahl (approved):
/lgtm

PR #439, opened 2023-02-10
Ensure conformance test's compatibility with ginkgo v1

hendrikKahl:

/area testing
/kind technical-debt

Ginkgo v1 is used by older k8s versions incl. v1.24. Only with v1.25 and later, k8s migrated everything to ginkgo v2. This leads to a situation, where we cannot run conformance tests with kubetest using ginkgo arguments that have been introduced with v2.

For now, I would opt to switch back to using deprecated ginkgo arguments. If those will be dropped in future while we still have to run conformance tests for k8s versions that come with ginkgo v1, we have determine the target cluster's version and set up the arguments accordingly.

Ref: https://github.com/kubernetes/kubernetes/commit/857458cfa5b0083bc55736aaccbcf1dc796ba320

Additionally, this PR contains an update to the latest version of `kubetest`.

Fixes #

/invite @dguendisch 

**Release note**:

```other operator
NONE
```

### DIFF
--- a/integration-tests/e2e/config/config.go
+++ b/integration-tests/e2e/config/config.go
@@ -71,7 +71,7 @@ var (
 
 const (
 	// points to a commit in https://github.com/kubernetes/test-infra/commits/master
-	TestInfraVersion = "590bc3fef25c06032c5cc53792c2a444a257d575"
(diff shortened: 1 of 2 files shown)


dguendisch (approved):
/lgtm

PR #440, opened 2023-02-15
Request and use a shoot/adminkubeconfig, if possible

### LINKED ISSUE #435: Add support for `adminKubeconfig` subresource to tests
As of today, `TestDefinitions` and their respective code in this repo expect a `kubeconfig` to be presented at runtime (https://github.com/gardener/test-infra/blob/master/.test-defs/conformanceTestgrid.yaml#L56). While this is still valid for local test executions, we should also support requesting/using short-lived kubeconfigs. 

Gardener issues short-lived admin kubeconfigs as an alternative to static tokens: https://gardener.cloud/docs/gardener/api-reference/authentication/#adminkubeconfigrequest

Ideally, the test framework used here can be adapted to:
- accept an existing kubeconfig, if it is presented with such a file
- request an `adminkubeconfig`, if a Gardener kubeconfig is presented

hendrikKahl:

/area testing
/kind test

In case a gardener project namespace, shoot name and gardener kubeconfig are specified to the invocation of `integration-tests/e2e`, a `shoot/adminkubeconfig` will be requested. The reply will be stored unter "/tmp/shoot.config" and all environment variables are changed accordingly. The kubeconfig will expire after 24h hours. All conformance tests running longer would be considered a failure anyways.

Otherwise the existing contract prevails where a valid kubeconfig is supplied by the invoking entity.

Fixes #435

/invite @dguendisch 
cc @ialidzhikov 

**Release note**:

```other operator
Allow conformance testframework to request a short-lived shoot/adminkubeconfig
```

### DIFF
--- a/integration-tests/e2e/kubetest/setup/kubeconfig.go
+++ b/integration-tests/e2e/kubetest/setup/kubeconfig.go
@@ -1,0 +1,76 @@
+// Copyright 2019 Copyright (c) 2023 SAP SE or an SAP affiliate company. All rights reserved. This file is licensed under the Apache Software License, v. 2 except as noted otherwise in the LICENSE file.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package setup
+
+import (
+	"context"
+	"os"
+
+	authenticationv1alpha1 "github.com/gardener/gardener/pkg/apis/authentication/v1alpha1"
+	"github.com/gardener/gardener/pkg/client/core/clientset/versioned/typed/core/v1beta1"
+	"github.com/gardener/test-infra/integration-tests/e2e/config"
+	log "github.com/sirupsen/logrus"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/tools/clientcmd"
+	"k8s.io/utils/pointer"
+)
+
+// tokenExpiresAfterSeconds sets the token to expire after 24h. All tests running longer will be considered as failed.
+const tokenExpiresAfterSeconds int64 = 86400
+
+// setupKubeconfig requests an admin kubeconfig for a given shoot
+func setupKubeconfig() error {
+
+	// check, if a shoots/adminkubeconfig can be requested
+	// see https://gardener.cloud/docs/gardener/usage/shoot_access/#shootsadminkubeconfig-subresource for details
+	if config.GardenKubeconfigPath != "" && config.ProjectNamespace != "" && config.ShootName != "" {
+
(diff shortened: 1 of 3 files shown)


integration-tests/e2e/kubetest/setup/kubeconfig.go
```diff
@@ -0,0 +1,76 @@
...
+		}
+		restConfig, err := gardenClientConfig.ClientConfig()
+		if err != nil {
+			return err
+		}
+		gardenerRestClient, err := v1beta1.NewForConfig(restConfig)
+		if err != nil {
+			return err
+		}
+		adminKubeconfigRequest, err = gardenerRestClient.Shoots(config.ProjectNamespace).CreateAdminKubeconfigRequest(context.Background(), config.ShootName, adminKubeconfigRequest, metav1.CreateOptions{})
```
ialidzhikov, line 62: Starting v0.14.x in the controller-runtime client supports subresources - see https://github.com/gardener/gardener/blob/8d53342ddcf0112209ce85662d7ba6bbcf2890aa/test/utils/shoots/access/adminkubeconfig.go#L44-L53

When you upgrade to controller-runtime@v0.14.x, you could consider using it.

ialidzhikov (approved):
/lgtm

ialidzhikov:
Okay, I now see that this change is rather related to the topic for which I already contacted you for advice in the beginning of January - the first item from the follow-up items in https://github.com/gardener/gardener/issues/6773. I was also planning these days to look again into it.
From the details I have so far I think we have 2 options:
1. Adapt the Shoot creation test in https://github.com/gardener/gardener/blob/7c63e2a6c5d46f9f1cb676602f864275ab8cab40/test/framework/shootcreationframework.go#L423-L425 to do not save the static token kubeconfig but to save a kubeconfig obtained from `shoot/adminkubeconfig` under `$TM_KUBECONFIG_PATH/shoot.config`.
2. Adapt usages of `$TM_KUBECONFIG_PATH/shoot.config` to do not rely on this downloaded kubeconfig but to request their own - as done in this PR. With this approach we no longer need `$TM_KUBECONFIG_PATH/shoot.config` as every test will be requesting its own kubeconfig.

I was thinking to tackle the problem by using 1. I guess if there are other usages of `$TM_KUBECONFIG_PATH/shoot.config` they also have implement a similar logic. WDYT about using option 1.?

ialidzhikov:
The required adaptation (ref https://github.com/ialidzhikov/gardener/commit/f696498782467d2bba5361e60f4251e4dfa915df) if we prefer to proceed with option 1.
I don't know why in my initial email from January I suggested option 2. as the only possible option and overlook option 1. I am sorry again for this!
Although, I am also happy to hear your opinion on which option/approach you prefer.

hendrikKahl:
Option 1 sounds reasonable to me especially as it would keep the "contract" of a shoot kubeconfig being available. I'm not aware of any tests with a runtime of +24h and they are still free to implement their own logic. Thanks for bringing up the topic!

To avoid some duplicate code in different repos, I would just close this PR. 

> I don't know why in my initial email from January I suggested option 2. as the only possible option and overlook option 1. I am sorry again for this!
Although, I am also happy to hear your opinion on which option/approach you prefer.

Absolutely nothing worry about :) I didn't think of it initially as well. 
Also, I refreshed my knowledge on clients for this PR but otherwise didn't spend too much time on it.

ialidzhikov:
Thank you @hendrikKahl for your help!
I created https://github.com/gardener/gardener/pull/7495, I will give it a test spin tomorrow morning and then promote to ready.

PR #444, opened 2023-03-28
Replace helm v2 dependency with helm v3

hendrikKahl:

/area testing
/kind technical-debt

`testrunner` has a dependency to the go modules of helm v2. It is used to render helm-style `TestRun` charts.

This PR replaces helm v2 with helm v3 (latest & greatest).

Fixes #

/invite @dguendisch 

**Release note**:

```other developer
`testrunner` uses helm v3 modules to render testrun charts.
```


dguendisch (approved):
/lgtm

PR #449, opened 2023-04-13
Cleanup missed todo's 

hendrikKahl:

/area testing
/kind cleanup

- Update the golang image used for testing, if the test definitions in this repo actually compile.
- adjust RBAC rule for the argo workflow executor. So far, it was still using the legacy approach of patching the pod. With v3.4.x however, the `workflowtaskresults` resource can be used for it ([ref](https://argoproj.github.io/argo-workflows/workflow-rbac/)).

Fixes #

/invite @dguendisch 

**Release note**:

```other operator
Use golang 1.20 for `build-test-defs.sh` and adjust argo wf executor role
```

### DIFF
--- a/charts/testmachinery/charts/argo/templates/rbac.yaml
+++ b/charts/testmachinery/charts/argo/templates/rbac.yaml
@@ -322,30 +322,19 @@ rules:
       - update
       - patch
       - delete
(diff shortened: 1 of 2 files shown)


dguendisch (approved):
/lgtm

PR #460, opened 2023-07-11
Make error msg more helpful

dguendisch:

/area usability
/kind cleanup

Make the error more helpful when the testrunner fails to find any non-expired K8s version because of some K8s pattern specified in a flavor.

/invite @hendrikKahl 
cc @rfranzke 

Fixes #

**Release note**:

```other operator

```

### DIFF
--- a/pkg/util/kubernetes_version.go
+++ b/pkg/util/kubernetes_version.go
@@ -65,7 +65,7 @@ func GetK8sVersions(cloudprofile gardencorev1beta1.CloudProfile, config common.S
 	}
 
 	if len(filtered) == 0 {
-		return nil, fmt.Errorf("no K8s version can be specified")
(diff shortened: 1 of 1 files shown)


pkg/util/kubernetes_version.go
```diff
@@ -65,7 +65,7 @@ func GetK8sVersions(cloudprofile gardencorev1beta1.CloudProfile, config common.S
 	}
 
 	if len(filtered) == 0 {
-		return nil, fmt.Errorf("no K8s version can be specified")
+		return nil, fmt.Errorf("no non-expired K8s version could be found in current cloudprofiles for testflavor pattern %s", pattern)
```
hendrikKahl, line 68:
```suggestion
		return nil, fmt.Errorf("no non-expired K8s version could be found in current cloudprofile %s for testflavor pattern %s", cloudprofile.Name, pattern)
```

hendrikKahl (approved):
/lgtm

🎉 🚀

PR #468, opened 2023-09-27
Don't panic on unexpected git url scheme but error out

dguendisch:

/area testing
/kind  bug

Not sure why exactly all these fields are parsed out from a presumable github.com URL but if it's in an unexpected format we should simply error out and thus skip a test location instead of panic'ing.

/invite @hendrikKahl 

Fixes #

**Release note**:

```other user
Fix a panic on unexpected git location URLs
```


hendrikKahl (commented):
/lgtm

PR #480, opened 2023-12-12
Filter for duplicate locations and take the greatest revision

hendrikKahl:

/area testing
/kind task

Enhance the function that parses components and adds them as locations to the testruns in a way, that it avoids adding duplicates. When a duplicate is found, the greatest version will be chosen ("main" or "master" tops semVer). The overwrite mechanism with additionalLocations will continue to function.

Fixes #

/invite @dguendisch 

**Release note**:

```other operator
Filter for duplicate locations and take the greatest revision
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -68,8 +68,8 @@ RUN  \
     rsync \
     bc \
     linux-headers \
-  && pip install google-crc32c \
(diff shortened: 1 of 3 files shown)


dguendisch (approved):
/lgtm

PR #486, opened 2023-12-21
Fix `make generate` in pipeline step `check`

hendrikKahl:

/area testing
/kind bug

Adapt the `check` step in our pipelines to create symlink of the repo root to `/go/src/github.com/gardener/test-infra`. This path is used by `make generate` (actually by https://github.com/kubernetes/code-generator/blob/master/generate-internal-groups.sh) to store the result.

Additionally, run `make generate` to get a clean state.

Fixes #

/invite @dguendisch 

cc @fabianburth 

**Release note**:

```other operator
NONE
```


dguendisch (approved):
/lgtm

PR #488, opened 2024-01-10
e2e: add support for skipping individual tests

hendrikKahl:

/area testing
/kind task

Sometimes, individual test might not work with all environments. This PR adds support to the e2e framework for skipping individual tests.

Tests may be skipped via the new cmd line flag or an environment variable. For example, a testrun based on the [conformance test definition](https://github.com/gardener/test-infra/blob/44f0f25525f5dee74d44421b12a6eda1a4b07c13/.test-defs/conformanceTestgridParallel.yaml) may skip a test by adding a config to a step like this:

```
config:
- name: SKIP_INDIVIDUAL_TEST_CASES
  type: env
   value: "HostPort.validates.that.there.is.no.conflict.between.pods.with.same.hostPort.but.different.hostIP.and.protocol"
```

The value is interpreted as a regex by ginkgo, hence escape spaces for example with `\s` would be possible too. To add skip multiple tests, everything needs to go into a single value with `|` as delimiter.

Fixes #

/invite @dguendisch 

**Release note**:

```other operator
NONE
```


dguendisch (approved):
/lgtm

PR #496, opened 2024-02-12
Drop vendor directory & update dependencies

hendrikKahl:

/area testing
/kind task

With this PR we would drop the vendor directory and instead use a virtual GOPATH (https://github.com/ironcore-dev/vgopath).

There are several adaptions to make scripts / make targets work with the new setup. Additionally, dependencies for Argo and Gardener (+a few others) have been bumped.

Fixes #

/invite @dguendisch 

**Release note**:

```other operator
NONE
```


dguendisch (approved):
/lgtm (after fixing the test-definitions script :))

PR #525, opened 2024-07-03
Add latest-x pattern for major machine image versions

hendrikKahl:

/area testing
/kind enhancement

Testrunner can parse latest-x pattern for major machine image and fetch the correct version from a cloudprofile. It respects expiration dates and architecture. Build versions continue to be ignored.

In case the pattern is not recognized , the version will be interpreted as string and validation is delegated to the respective garden cluster during shoot creation.

Fixes #

/invite @dguendisch 

**Release note**:

```other operator
Latest-x pattern can be used for machine image versions in testrun flavors
```

### DIFF
--- a/pkg/util/machine_image_test.go
+++ b/pkg/util/machine_image_test.go
@@ -1,10 +1,76 @@
 package util
 
 import (
+	"time"
+
 	gardencorev1beta1 "github.com/gardener/gardener/pkg/apis/core/v1beta1"
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"github.com/gardener/test-infra/pkg/common"
 )
+
+const (
+	imageName  = "gardenlinux"
+	arch_amd64 = "amd64"
+	arch_arm64 = "arm64"
+)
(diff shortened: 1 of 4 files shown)


pkg/util/machine_image_test.go
```diff
@@ -1,11 +1,77 @@
...
+				MachineImages: []gardencorev1beta1.MachineImage{
+					{
+						Name: imageName,
+						Versions: []gardencorev1beta1.MachineImageVersion{
+							{ExpirableVersion: gardencorev1beta1.ExpirableVersion{
+								Version:        "3.4.5",
+								ExpirationDate: &futureTime,
+							}, Architectures: []string{arch_amd64, arch_arm64}},
+							{ExpirableVersion: gardencorev1beta1.ExpirableVersion{
+								Version:        "2.3.4",
```
dguendisch, line 41: Maybe also add a version `2.3.3` so that we test that n-1 would in fact return `2.3.4` instead of `2.3.3`?

hendrikKahl: https://github.com/gardener/test-infra/pull/525/commits/d03e116b02fe86dedfab749594d0afc4215ac45a

dguendisch (approved):
/lgtm

PR #532, opened 2024-09-04
Ocm and cleanup

ccwienk:
We plan to keep publishing to PYPI, however at a reduced cadence (e.g. every other month / quarter), and leverage OCM as primary means of distributing our python-packages instead. Therefore, prepare dockerfile accordingly.

Also, prepare for announced expiry change of CAs + add successor CA already.


hendrikKahl:
@ccwienk thanks for the PR 🚀 
I noticed `ghcr.io/open-component-model/ocm/ocm.software/ocmcli/ocmcli-image` is only available for `amd64`. Are there any plans to publish it for `arm64` as well? It would make the local development scenario a bit easier :) Thanks!

hendrikKahl (approved):
/lgtm

PR #551, opened 2024-12-06
`prepare` clones with `--filter=blob:none`

hendrikKahl:

/area testing
/kind enhancement

Currently, the `prepare` step clones full repositories. To fetch less data, the `--filter=blob:none` can be used. While `--filter=tree:0` would be an alternative, it might have issues with submodules. 

Fixes #

/invite @dguendisch 

**Release note**:

```other operator
`prepare` clones with `--filter=blob:none`
```


dguendisch (approved):
/lgtm

PR #603, opened 2025-02-25
Update go to 1.24 and adjust tooling

hendrikKahl:

/area testing
/kind task

Update testmachinery to use Go 1.24 and the new tools directive. The code generators imported through gardener/gardener do not use 1.24 yet, which is why we need to build them ourselves.

Additionally, I updated go dependencies for gardener/gardener and argo. 

Fixes #

/invite @dguendisch 

**Release note**:

```other operator
NONE
```


dguendisch:
> The code generators imported through gardener/gardener do not use 1.24 yet, which is why we need to build them ourselves.

So we'll remove the building ourselves once g/g uses 1.24?

hendrikKahl:
> > The code generators imported through gardener/gardener do not use 1.24 yet, which is why we need to build them ourselves.
> 
> So we'll remove the building ourselves once g/g uses 1.24?

Yep, that's the plan :) in the meantime it doesn't hurt to run the build command as part of our makefile.

dguendisch (approved):
/lgtm

PR #619, opened 2025-03-11
Update ingress name for lookup

hendrikKahl:

/area testing
/kind bug

Update the monitoring `ingress` name.

Fixes #

/invite @dguendisch 

**Release note**:

```other operator
NONE
```

### DIFF
--- a/pkg/util/util.go
+++ b/pkg/util/util.go
@@ -481,15 +481,15 @@ type ESHits struct {
 	} `json:"hits"`
 }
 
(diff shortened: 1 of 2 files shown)


dguendisch:
Thank you!
While it won't fail (because the test is at a lower level), maybe also change https://github.com/gardener/test-infra/blob/master/pkg/util/util_test.go#L29

hendrikKahl:
> Thank you! While it won't fail (because the test is at a lower level), maybe also change https://github.com/gardener/test-infra/blob/master/pkg/util/util_test.go#L29

Done :) Thanks for spotting it

PR #642, opened 2025-03-31
Fix argo configmap

hendrikKahl:

/area testing
/kind bug

Apparently, the workflow controller expect yaml and not a multi-line string to parse default.

Fixes #

/invite @dguendisch 

**Release note**:

```other operator
NONE
```


dguendisch (approved):
/lgtm

PR #736, opened 2025-07-29
Add functionality to support machine image update tests

acumino:

/area testing
/kind enhancement

This PR adds functionality to support machine image update tests, which are only relevant only if shoot has worker pool with an in-place update strategy.
This PR introduces a new value `machineImagePrevVersion`, which is the latest previous version from which a machine image can be in-place updated to the current machine image version.
It also sets the value `shoot.machine.imageversion`, which is the version the machine image will be updated to from the `machineImagePrevVersion` version.

Part of https://github.com/gardener/gardener/issues/10219

/invite @hendrikKahl @dguendisch 

**Release note**:

```other operator
NONE
```


hendrikKahl (approved):
/lgtm

@acumino thanks for the PR :) could you please rebase? Thanks!

hendrikKahl (approved):
/lgtm

PR #786, opened 2025-09-30
Testruner posts to GitHub actions step summary file

hendrikKahl:

/area testing
/kind task

Testrunner has a new flag to make it post updates to a GitHub Actions step summary file.
During an action run it will post:
- a link to the execution group
- links to each testrun
- result of each testrun

When all tests have finished, it will post:
- a link to the execution group
- results of all testruns

Fixes #

/invite @dguendisch 

**Release note**:

```other operator
Testruner posts to GitHub actions step summary file
```

### DIFF
--- a/cmd/testrunner/cmd/run_template/options.go
+++ b/cmd/testrunner/cmd/run_template/options.go
@@ -43,6 +43,7 @@ type options struct {
 	filterPatchVersions      bool
 	failOnError              bool
 	timeout                  int64
+	postToGitHubStepSummary  string
 }
 
 // NewOptions creates a new options struct.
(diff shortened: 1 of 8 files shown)


hendrikKahl:
> /lgtm although we could also drop the configurable path as the GHA file is anyway specified to be configured via env variable `$GITHUB_STEP_SUMMARY` (https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-commands#adding-a-job-summary), but having the file location configurable is also fine.

Thanks for the suggestion :)

I prefer to leave it to the user/invoker of `testrunner` to decide whether or not it should post to a file.

dguendisch:
Slight misunderstanding: I was just proposing to not make the path configurable. I of course agree to make writing to `$GITHUB_STEP_SUMMARY` optional/configurable. But anyway, no strong opinion on the path itself, so all good 😊

PR #928, opened 2026-03-11
add ingressClassName to Argo + TM-bot ingress templates

IndritFejza:
We might need to specify directly in the ingress spec what ingress class is used.

If we have 2 ingress controllers in a cluster, the ingress resources needs to have it specifies which controller is responsible for them.

/area testing
/kind task

Fixes #

**Release note**:

```other operator
NONE
```


dguendisch (approved):
/lgtm

PR #963, opened 2026-04-22
add retries to status uploader

IndritFejza:
add some retries to status-uploader so we don't fail with only 1 try

/area testing
/kind task

Fixes #

**Release note**:

```other operator
add retries to status-uploader 
```


IndritFejza:
will follow up on the logic added here: https://github.com/gardener/test-infra/pull/941 not with retries.

PR #967, opened 2026-04-24
add annotation support for argo-server svc

IndritFejza:
add annotation support for the argo-server Service. 
Some ingress controllers like traefik, sometimes require services to be annotated. 
An example annotation would be `"traefik.ingress.kubernetes.io/service.serversscheme=https"` and this needs to be added to the target-service, [see docs](https://doc.traefik.io/traefik/reference/routing-configuration/kubernetes/ingress/#on-service). 

Other ingress controllers, like nginx would only require ingresses to be annotated, not services, like [here](https://github.com/gardener/test-infra/blob/a1039dc8fe7cefc57e0141988922394648a14422/charts/testmachinery/charts/argo/templates/ingress.yaml#L15)!

/area testing
/kind task

Fixes #

**Release note**:

```other operator
add possibility to add annotations to the argo-server service
```

### DIFF
--- a/charts/testmachinery/charts/argo/templates/argo.yaml
+++ b/charts/testmachinery/charts/argo/templates/argo.yaml
@@ -8,10 +8,16 @@ kind: Service
 metadata:
   name: argo-server
   namespace: {{.Release.Namespace}}
+  {{- if .Values.argoserver.service.annotations }}
+  annotations:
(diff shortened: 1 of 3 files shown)


dguendisch (commented):
/lgtm
/approve

dguendisch (approved):

PR #984, opened 2026-06-01
add support for machineCapabilities in CloudProfile spec

IndritFejza:
Based on GEP33: https://github.com/gardener/gardener/issues/11301 /  https://github.com/gardener/gardener-extension-provider-aws/pull/1645 we need to add support here so the architecture can be parsed correctly.

/area testing

Fixes #

**Release note**:

```other operator
Add parsing support fro architecture from cp.spec.machineCapabilities
```

/kind task

### DIFF
--- a/pkg/util/machine_image.go
+++ b/pkg/util/machine_image.go
@@ -8,6 +8,7 @@ import (
 
 	"github.com/Masterminds/semver/v3"
 	gardencorev1beta1 "github.com/gardener/gardener/pkg/apis/core/v1beta1"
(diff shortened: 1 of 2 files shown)


hendrikKahl (approved):
/lgtm

Roncossek:
If the cloudprofile is read from an kube-api-server this should not be required. 
for now: 
- even with machineCapabilities: the architectures field should be populated

for long term:
- the architectures field will be deprecated and `capabilityFlavors[].architecute` should be the place to look out. 

--> versions are read from file system. so the above is not applicable.

IndritFejza:
see: https://github.com/gardener/test-infra/pull/986 for up to date changes

PR #986, opened 2026-06-09
Derive machine image architectures from cloudprofile providerConfig

Roncossek:

/area testing
/kind bug

The `state` cloudprofile generators stopped populating the legacy `MachineImageVersion.Architectures` field, so `FilterArchSpecificMachineImage` was filtering every version out for any architecture. This rewrites the filter to read architecture data from `cloudprofile.Spec.ProviderConfig` (region, flat, and capability-flavor formats) and to honor a single-valued `Spec.MachineCapabilities[architecture]` as an implicit cloudprofile-wide arch. Region and flat formats are isolated behind a labeled section so they can be removed once provider migration to capability flavors completes.

Fixes #

`FilterArchSpecificMachineImage` was renamed and resignatured to `FilterMachineImageVersionsByArch(cloudprofile, imageName, versions, architecture)`.

Alternative approach to: https://github.com/gardener/test-infra/pull/984

**Release note**:
```bugfix operator
Machine image architectures are now read from `CloudProfile.spec.providerConfig` (and `spec.machineCapabilities`) instead of the unset legacy `MachineImageVersion.architectures` field, restoring architecture-based version filtering for shoot creation.
```

### DIFF
--- a/pkg/shootflavors/extendedflavors_test.go
+++ b/pkg/shootflavors/extendedflavors_test.go
@@ -19,6 +19,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	"github.com/gardener/test-infra/pkg/common"
+	"github.com/gardener/test-infra/pkg/util"
 )
 
 var _ = Describe("extended flavor test", func() {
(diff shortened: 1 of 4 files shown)


IndritFejza (approved):
/lgtm
approve

Tests using this Snapshot version are green.

PR #1010, opened 2026-08-04
testrunner: surface CloudProfile path errors instead of silently exiting

matthias-horne:
/area testing
/kind bug

When `--cloud-profile-search-path` points to a non-existent or empty directory, `Complete()` calls `logger.Log.Error(...)` followed by `os.Exit(1)`. However, `logger.Log` is a zero-value `logr.Logger` (a discard sink since logr v1.0) at that point, so no output is produced and the process exits silently with code 1.

The fix returns errors from `Complete()` instead, which bubble up to the `Run` cobra handler where they are printed via `fmt.Println`.

Fixes #

Verified locally:
- Non-existent path: `unable to find CloudProfiles in files at /nonexistent/path: CloudProfile search path /nonexistent/path does not exist: stat /nonexistent/path: no such file or directory`
- Existing but empty directory: `no CloudProfiles found in files at /tmp/empty-cloudprofiles`
- Valid path: `Complete()` succeeds, testrunner proceeds normally

**Release note**:
```bugfix developer
testrunner now prints an error message when --cloud-profile-search-path is invalid or empty, instead of exiting silently with code 1.
```

### DIFF
--- a/cmd/testrunner/cmd/run_template/options.go
+++ b/cmd/testrunner/cmd/run_template/options.go
@@ -6,6 +6,7 @@ package run_template
 
 import (
 	"errors"
+	"fmt"
 	"os"
 	"path/filepath"
 	"time"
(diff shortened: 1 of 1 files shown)


IndritFejza (approved):
/lgtm
/approve

